PR #57, opened 2026-01-28
feat(testing): add mcp integration tests

NeethuESim:
### What are the steps to test this PR contribution?

Prerequisites

- Build the mcp server docker image 
- Install the required test packages using - `pip install -r tests/requirements.txt`

How does a reviewer know it works?

- Run the test script - `python -m pytest -s tests/test_mcp.py`
- Check if following 2 docker containers have started - mcp server & testcontainer 
- All tests should pass without any errors. Warnings can be ignored.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -23,7 +23,7 @@ jobs:
           pip install -r mcp-local/tests/requirements.txt
 
       - name: Build MCP Docker image
-        run: docker build -t arm-mcp:latest -f mcp-local/Dockerfile mcp-local
+        run: docker buildx build -f mcp-local/Dockerfile -t arm-mcp mcp-local
 
       - name: Run integration tests
         env:
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -23,7 +23,7 @@ jobs:
           pip install -r mcp-local/tests/requirements.txt
 
       - name: Build MCP Docker image
-        run: docker buildx build -f mcp-local/Dockerfile -t arm-mcp mcp-local
+        run: docker buildx build -f mcp-local/Dockerfile -t arm-mcp .
 
       - name: Run integration tests
         env:
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -23,7 +23,7 @@ jobs:
           pip install -r mcp-local/tests/requirements.txt
 
       - name: Build MCP Docker image
-        run: docker buildx build --platform linux/arm64,linux/amd64 -f mcp-local/Dockerfile -t arm-mcp .
+        run:  docker buildx build -f mcp-local/Dockerfile -t arm-mcp .
 
       - name: Run integration tests
         env:
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -23,7 +23,7 @@ jobs:
           pip install -r mcp-local/tests/requirements.txt
 
       - name: Build MCP Docker image
-        run: docker buildx build -f mcp-local/Dockerfile -t arm-mcp .
+        run: docker buildx build --platform linux/arm64,linux/amd64 -f mcp-local/Dockerfile -t arm-mcp .
 
       - name: Run integration tests
         env:
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   integration-tests:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - name: Checkout
         uses: actions/checkout@v4
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,0 +1,31 @@
+name: Integration Tests
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+          cache: pip
+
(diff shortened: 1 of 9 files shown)


mcp-local/tests/test_mcp.py
```diff
@@ -0,0 +1,158 @@
...
+    while len(header) < 8:
+        if time.time() > deadline:
+            raise TimeoutError("Timed out waiting for docker frame header.")
+        chunk = sock.recv(8 - len(header))
+        if not chunk:
+            time.sleep(0.01)
+            continue
+        header += chunk
+
+    if header[1:4] != b"\x00\x00\x00":
```
JoeStech, line 48: What is the intention for this? It looks like it might silently return malformed data as if it were valid. Please add a comment clarifying this behavior.

README.md
```diff
@@ -119,6 +119,19 @@ After updating the configuration, restart your MCP client to load the Arm MCP se
   - `Dockerfile`: Multi-stage Docker build
 - **`embedding-generation/`**: Scripts for regenerating the knowledge base from source documents
 
+## Integration Testing
+
+### Pre-requisites
+
+- Build the mcp server docker image
+- Install the required test packages using - `pip install -r tests/requirements.txt`
```
JoeStech: please clarify that this is in the mcp_local directory.

mcp-local/tests/constants.py
```diff
@@ -0,0 +1,237 @@
...
+                "arguments": {
+                    "image": "armswdev/arm-mcp",
+                    "invocation_reason": (
+                        "Checking the architecture support of the armswdev/arm-mcp container image to verify ARM compatibility as requested by the user."
+                    ),
+                },
+            },
+        }
+
+EXPECTED_CHECK_SKOPEO_RESPONSE = {
```
JoeStech, line 78: please remove timestamps and the Digest hashes from this structure, and only assert on stable fields like status, architecture, and OS. Otherwise this test will immediately break when the upstream image is tweaked in any way.

mcp-local/tests/constants.py
```diff
@@ -0,0 +1,237 @@
...
+            "id": 5,
+            "method": "tools/call",
+            "params": {
+                "name": "migrate_ease_scan",
+                "arguments": {
+                    "scanner": "java",
+                },
+            },
+        }
+EXPECTED_CHECK_MIGRATE_EASE_TOOL_RESPONSE = {
```
JoeStech, line 124: looks like the only part of this structure that is used is "status", can you please remove all the unused parts and put a TODO comment here that we need to do more extensive checking with a static example down the road? i.e. we'll want to create a repo that we control that we will be sure will be unchanging so we can check more detailed response params.

mcp-local/tests/test_mcp.py
```diff
@@ -0,0 +1,158 @@
...
+        #Check Skopeo Tool Test
+        raw_socket.sendall(_encode_mcp_message(constants.CHECK_SKOPEO_REQUEST))
+        check_skopeo_response = _read_response(3, timeout=60)
+        assert check_skopeo_response.get("result")["structuredContent"] == constants.EXPECTED_CHECK_SKOPEO_RESPONSE, "Test Failed: MCP check_skopeo tool failed: content mismatch. Expected: {}, Received: {}".format(json.dumps(constants.EXPECTED_CHECK_SKOPEO_RESPONSE,indent=2), json.dumps(check_skopeo_response.get("result")["structuredContent"],indent=2))
+        print("\n***Test Passed: MCP check_skopeo tool succeeded")
+
+        #Check NGINX Query Test
+        raw_socket.sendall(_encode_mcp_message(constants.CHECK_NGINX_REQUEST))
+        check_nginx_response = _read_response(4, timeout=60)
+        assert any(item in str(check_nginx_response.get("result")["structuredContent"]) for item in constants.EXPECTED_CHECK_NGINX_RESPONSE), "Test Failed: MCP check_nginx tool failed: content mismatch., Expected one of: {}, Received: {}".format(json.dumps(constants.EXPECTED_CHECK_NGINX_RESPONSE,indent=2), json.dumps(check_nginx_response.get("result")["structuredContent"],indent=2))
```
JoeStech: Converting structured content to a string and checking substrings is fragile. Please iterate like this:
  urls = check_nginx_response["result"]["structuredContent"]
  assert any(expected in urls for expected in EXPECTED_CHECK_NGINX_RESPONSE)

.github/workflows/integration-tests.yml
JoeStech, line 1: This fails: https://github.com/arm/mcp/actions/runs/21604768057/job/62382523919

Please look into this failure and set up a meeting with me if you need a second pair of eyes.

JoeStech (changes requested):

.github/workflows/integration-tests.yml
NeethuESim: > This fails: https://github.com/arm/mcp/actions/runs/21604768057/job/62382523919
> 
> Please look into this failure and set up a meeting with me if you need a second pair of eyes.

Yes, for this one I am waiting for github actions to be back up - https://www.githubstatus.com/

mcp-local/tests/constants.py
NeethuESim: @JoeStech should I remove the entire response & just keep status, architecture & OS fields for comparing between the actual & expected results ? This will make the code cleaner, however will loose details. I would prefer keeping the entire response (with digest/timestamps as empty fields) since it can help with debugging.

JoeStech: I'm fine keeping it in the file, ideally with a comment mentioning which fields are actually used by the test

mcp-local/tests/constants.py
NeethuESim: Added the comment, I am also leaving the response within comments for future reference. It can be removed later.

mcp-local/tests/test_mcp.py
NeethuESim: So this doesn't work because of how `in` operator works, it checks for exact membership when urls is list/dict. I will use json.dumps instead of str for more safer consistent conversion.

JoeStech (approved):

PR #61, opened 2026-02-12
feat: GitHub Action workflow for multi architecture build & integration testing on arm64 & amd64

NeethuESim:
How does a reviewer know it works?

- GitHub Action workflow should work for both arm64 & amd64 runners


.github/workflows/integration-tests.yml
```diff
@@ -6,8 +6,21 @@ on:
...
+      fail-fast: false
+      matrix:
+        include:
+          - runner: ubuntu-24.04
+            platform: linux/amd64
+          - runner: ubuntu-24.04-arm
+            platform: linux/arm64
+    runs-on: ${{ matrix.runner }}
     steps:
+      - name: Free disk space
```
JoeStech: this section can be removed now

mcp-local/tests/constants.py
```diff
@@ -12,7 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-MCP_DOCKER_IMAGE = "arm-mcp:latest"
+MCP_DOCKER_IMAGE_ARM = "arm-mcp:latest"
```
JoeStech: this doesn't seem consistent. You've got no armlimited here, but you do in the x86 variable below, but neither of these images should be arm or x86 specific.

JoeStech (changes requested):

mcp-local/tests/constants.py
NeethuESim: So when I tested earlier, the image names were different when built on arm (no armlimited) & x86 (with armlimited). Now I see the same name for both armlimited/arm-mcp:latest. Will try testing integration test with this now.

JoeStech (approved):

mcp-local/tests/test_mcp.py
```diff
@@ -160,11 +163,14 @@ def _read_response(expected_id: int, timeout: float = 10.0) -> dict:
...
-        print("\n***Test Passed: MCP mca tool succeeded")
+        #Check MCA Tool Test - works only on platform=linux/arm64
+        if platform == constants.DEFAULT_PLATFORM:
+            raw_socket.sendall(_encode_mcp_message(constants.CHECK_MCA_TOOL_REQUEST))
+            check_mca_response = _read_response(7, timeout=60)
+            assert check_mca_response.get("result")["structuredContent"]["status"] == constants.EXPECTED_CHECK_MCA_TOOL_RESPONSE_STATUS, "Test Failed: MCP mca tool failed: status mismatch.Expected: {}, Received: {}".format(json.dumps(constants.EXPECTED_CHECK_MCA_TOOL_RESPONSE_STATUS,indent=2), json.dumps(check_mca_response.get("result")["structuredContent"]["status"],indent=2))
+            print("\n***Test Passed: MCP mca tool succeeded")
+        else:
+            print("\n***Test NA: MCP mca tool is not supported on this platform: {}".format(platform))
```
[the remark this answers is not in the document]
NeethuESim: this will not print the reason for skipping the testcase.

PR #82, opened 2026-05-06
Add Java CI/CD integration test for Performix.

jaidev17:
- Prepare and compile CpuBurner on the runner for the APX Java hotspot test.
  - Simplify Java APX workload command in mcp-local/tests/constants.py.
  - Add success criteria: CpuBurner::workC must appear in Java hotspots.
  - Improve APX run_id parsing to handle multi-line JSON output.

### DIFF
--- a/mcp-local/tests/constants.py
+++ b/mcp-local/tests/constants.py
@@ -252,9 +252,8 @@ CHECK_APX_CPU_HOTSPOTS_JAVA_REQUEST = {
                 "name": "apx_recipe_run",
                 "arguments": {
                     "cmd": (
-                        "which java || (sudo apt-get update -qq && sudo apt-get install -y --no-install-recommends default-jdk-headless); "
                         "D=$(mktemp -d); "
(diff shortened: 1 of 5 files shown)


mcp-local/tests/constants.py
```diff
@@ -244,6 +244,22 @@
...
 EXPECTED_CHECK_MCA_TOOL_RESPONSE_STATUS = "ok"     
 
+CHECK_APX_CPU_HOTSPOTS_JAVA_REQUEST = {
+            "jsonrpc": "2.0",
+            "id": 9,
+            "method": "tools/call",
+            "params": {
+                "name": "apx_recipe_run",
+                "arguments": {
+                    "cmd": "java -XX:+PreserveFramePointer -cp /home/apxci/cpuburner CpuBurner 30",
```
JoeStech: is there a reason this needs to run for 30 seconds? We should limit the runtime of the tests wherever possible.

JoeStech (changes requested):

mcp-local/tests/constants.py
jaidev17: It is the time the CpuBurner loop runs for. We can reduce it as we are not checking the raw time numbers. I did not change it as 30 secs is the time used in the Golden Testcases that the performix team provided for testing. I can reduce it down to 10 secs, if that seems feasible.

JoeStech (approved):

PR #85, opened 2026-05-15
Extract shared knowledge base search and evaluation logic to submodule

brikin01:


embedding-generation/evaluate_retrieval.py
```diff
@@ -0,0 +1,79 @@
...
+        return 1
+
+    embedding_model = SentenceTransformer(
+        model_name,
+        cache_folder=sentence_transformer_cache_folder(),
+        local_files_only=True,
+    )
+    usearch_index = load_usearch_index(
+        str(index_path),
+        embedding_model.get_embedding_dimension(),
```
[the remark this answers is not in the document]
brikin01: This is an outdated suggestion, SentenceTransformer now uses `get_embedding_dimension` as seen in the [latest docs](https://www.sbert.net/docs/package_reference/sentence_transformer/model.html#sentence_transformers.sentence_transformer.model.SentenceTransformer.get_embedding_dimension). I have pinned the minimal version of SentenceTransformers in the pyproject.toml and the requirements.txt under `embedding_generation`, which is the only code that invokes this functionality.

embedding-generation/requirements.txt
```diff
@@ -3,6 +3,6 @@ beautifulsoup4
 pyyaml
 usearch
 boto3
-sentence-transformers
+sentence-transformers>=5.4
```
JoeStech, line 6: this is a good change, can you pin this version in the mcp-local requirements.txt also so they match?

embedding-generation/evaluate_retrieval.py
```diff
@@ -0,0 +1,79 @@
+"""Run a retrieval evaluation over the local metadata and index."""
```
JoeStech, line 1: is there a reason this filename was changed? I think it's probably fine, it just prevents a nice diff in the PR showing the actual lines that were changed

JoeStech (changes requested):

embedding-generation/requirements.txt
brikin01: Addressed in 2c381346622c9fcfa61c6b8b1fa026b55380a80b

embedding-generation/evaluate_retrieval.py
brikin01: I originally changed it to distinguish from a file evaluating the endpoint logic, but since that now will live in the endpoint repo itself, I've changed the name back as part of 2c381346622c9fcfa61c6b8b1fa026b55380a80b

brikin01:
@JoeStech Addressed the comments and also updated the evaluation function to output both the summary and actually include the individual cases so that we can do more detailed breakdowns later.

JoeStech (approved):

PR #86, opened 2026-06-03
Add UTM source to KB result links

jaidev17:
Adds utm_source=arm-mcp to all KB search result URLs.

Examples:
https://learn.arm.com/migration
→ https://learn.arm.com/migration?utm_source=arm-mcp

https://www.arm.com/developer-hub/ecosystem-dashboard/?package=curl
→ https://www.arm.com/developer-hub/ecosystem-dashboard/?package=curl&utm_source=arm-mcp

Testing done:
Built local Docker image arm-mcp:utm-test and verified real KB search results include utm_source=arm-mcp across plain URLs and URLs with existing query params.

### DIFF
--- a/arm_kb_search/resources.py
+++ b/arm_kb_search/resources.py
@@ -22,7 +22,7 @@ from usearch.index import Index
 
 from .config import K_RESULTS
 from .loaders import load_metadata, load_usearch_index
-from .response import add_disclaimer_to_arm_results
+from .response import add_disclaimer_to_arm_results, add_utm_source_to_results
 from .search import build_bm25_index, deduplicate_urls, hybrid_search
 
 
@@ -34,6 +34,7 @@ class SearchResources:
     bm25_index: BM25Okapi | None
     default_k: int = K_RESULTS
     include_disclaimers: bool = True
+    utm_source: str | None = None
 
 
 def sentence_transformer_cache_folder() -> str | None:
@@ -76,6 +77,7 @@ def load_search_resources(
     local_files_only_first: bool = True,
     default_k: int = K_RESULTS,
     include_disclaimers: bool = True,
+    utm_source: str | None = None,
 ) -> SearchResources:
     metadata = load_metadata(metadata_path)
     embedding_model = load_embedding_model(
@@ -95,6 +97,7 @@ def load_search_resources(
         bm25_index=bm25_index,
         default_k=default_k,
         include_disclaimers=include_disclaimers,
+        utm_source=utm_source,
     )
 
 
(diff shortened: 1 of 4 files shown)


arm_kb_search/response.py
```diff
@@ -42,3 +45,27 @@ def add_disclaimer_to_arm_results(results: List[Dict[str, Any]]) -> List[Dict[st
...
+    query_params.append(("utm_source", utm_source))
+    return urlunparse(parsed._replace(query=urlencode(query_params)))
+
+
+def add_utm_source_to_results(
+    results: List[Dict[str, Any]],
+    utm_source: str = ARM_MCP_UTM_SOURCE,
+) -> List[Dict[str, Any]]:
+    return [
+        {**item, "url": add_utm_source_to_url(item.get("url"), utm_source)} if "url" in item else item
```
brikin01, line 69: Did we want to append `utm_source` to all URLs in the knowledge base, or only Arm-owned URLs? Appending it to non-Arm URLs is probably harmless, but since the KB includes external sources too, I wanted to check that this is intentional.

arm_kb_search/resources.py
```diff
@@ -126,6 +126,7 @@ def search(
         }
         for item in deduped
     ]
+    formatted = add_utm_source_to_results(formatted)
```
brikin01: Could we avoid hard-coding/defaulting to `arm-mcp` inside the shared `arm_kb_search` path? Since this library is also used by the knowledge-base endpoint, it would be better if the caller supplied the `utm_source` value, with the library defaulting to no UTM source.

My suggested approach is for [`SearchResources`](https://github.com/arm/mcp/blob/73b071504448d87443b2f193404657082705bfe6/arm_kb_search/resources.py#L30-L36) to carry an optional `utm_source`; MCP would pass `arm-mcp` when loading resources, and the endpoint would pass its own suffix.

Open to other approaches, but I think the key point is that the shared library should not assume the MCP source.

brikin01 (changes requested):
Thanks for these changes! One question/clarification and one requested change.

arm_kb_search/response.py
jaidev17: Yes, confirmed with Joe that we want it in non Arm-owned URLs as well. We would like other sources to see that traffic is being sourced through the Arm MCP server.

arm_kb_search/resources.py
jaidev17: Good point! I made the changes.

brikin01 (approved):
Thanks for the changes and for double-checking with Joe! LGTM!

PR #87, opened 2026-06-04
Selectively include developer.arm.com content

apickard:
Uses the developer.arm.com search api to retrieve search results, filters them to just the relevant ones, and then generates embeddings from them.


JoeStech:
I like your general approach, my main request is to pull the changes to generate-chunks.py into a script that can be run on one-time basis to add values to the vector-db-sources.csv file.

The idea being that we want to be able to identify what pages are being added to the sources CSV without re-running it sight unseen during the weekly build, so that we can also add questions and expected URLs to the eval_questions.json file each time the script is run.

Let me know if you have any questions about this, or want to chat about it!

apickard:
This pull request will be superseded by a later one.

PR #88, opened 2026-06-09
feat: add new image tags, git tags, dry run for testing

NeethuESim:
[STESOL-507](https://jira.arm.com/browse/STESOL-507) - Arm MCP Server environment versioning improvements

## Docker image tags
The verified multi-arch image is published to Docker Hub with:
- `latest` — the **prod** image.
- A semantic version tag `X.Y.Z` (e.g. `2.3.0`), auto-computed (not hardcoded). The highest existing `X.Y.Z` tag from git version is the source of truth; the next version is derived from it:
  - **minor** (`X.Y+1.0`) — automatic on weekly knowledge-base rebuilds (`workflow_run`).
  - **major** (`X+1.0.0`) / **hotfix** (`X.Y.Z+1`) — selected manually via the `bump_type` input.
- An immutable dated tag `YYYY-MM-DD-github.run_number` (e.g. `2026-06-15-1234`) — the **stage** image.

`latest`/version aliases are applied only after the dated manifest is created and verified to contain both `linux/amd64` and `linux/arm64`.

## Git tags & release
- Pushes a matching annotated git tag `vX.Y.Z` for each published Docker version.
- Creates a GitHub Release with an auto-generated changelog (`.github/release.yml`).

## server.json
- Bump `mcp-local/server.json` (`version` + OCI package `identifier`) to the computed version & directly merge into the main.

## Safety / testing
- `dry_run` flag builds/validates without pushing images, manifests, git tags, or releases.
- Publishing is gated to `main`; non-main runs must use `dry_run=true` (fail-fast guard).


.github/workflows/build-mcp-image.yml
```diff
@@ -77,11 +87,44 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - name: Determine version tag
+        id: version
+        run: |
+          TAG_VERSION="2.2.0"
+          echo "tag_version=${TAG_VERSION}" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
NeethuESim: Updating it manually for now.

brikin01:
I think we might also need to update mcp-local/server.json with the latest version tag?

.github/workflows/build-mcp-image.yml
```diff
@@ -77,11 +99,55 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - name: Determine version tag
+        id: version
+        run: |
+          TAG_VERSION="2.2.0"
```
JoeStech: This should auto-increment based on the latest version number.

.github/workflows/build-mcp-image.yml
```diff
@@ -5,6 +5,11 @@ on:
     workflows: ["Build Embeddings Image"]
     types: [completed]
   workflow_dispatch:
```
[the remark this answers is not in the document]
JoeStech: this is a good idea, we want to gate on main

.github/workflows/build-mcp-image.yml
JoeStech, line 1: In this file we also need an update to the changelog (which doesn't yet exist, it needs to be added) denoting that the auto-deployed version is a weekly knowledge base update

.github/workflows/build-mcp-image.yml
```diff
@@ -77,11 +99,55 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
 
+      - name: Determine version tag
+        id: version
+        run: |
+          TAG_VERSION="2.2.0"
+          echo "tag_version=${TAG_VERSION}" >> "$GITHUB_OUTPUT"
+
       - name: Create multi-arch manifest for date tag
```
JoeStech, line 172: adding to this existing build step is a little bit brittle, we don't want to tag the version number + latest until we're sure the arch manifests are merged correctly. Please move the version number + latest tagging into the git tagging section, and leave this build step as-is.

So the flow should be:

  1. Build and push arch images.
  2. Create the date-tagged multi-arch manifest only.
  3. Verify/inspect the date tag.
  4. Compute the next 2.X.0.
  5. Add latest and 2.X.0 as aliases of the verified date tag.
  6. Create the matching git tag and changelog/release notes.

JoeStech (changes requested):

.github/workflows/build-mcp-image.yml
brikin01: Related to my above comment, should we also add updating the versions in mcp-local/server.json as a step between 4 and 5? It seems some upstream locations like the docker mcp registry (https://hub.docker.com/mcp/server/arm-mcp/overview) might be using the version listed in that file as its 'source of truth'? Please correct me if I'm wrong

NeethuESim: I wasn't sure if these need to be updated manually or automated. Will wait for Joe's response.

.github/workflows/build-mcp-image.yml
```diff
@@ -64,24 +123,406 @@ jobs:
...
+        # git tag, and server.json steps are all is_main-gated), so skipping it
+        # there avoids needlessly hitting (and possibly being rate-limited by)
+        # the Docker Hub API on validation runs.
+        if: ${{ steps.gate.outputs.is_main == 'true' }}
+        env:
+          # Minor bumps are automatic (workflow_run knowledge-base rebuilds).
+          # Manual runs must choose major or hotfix via the bump_type input.
+          BUMP_TYPE: ${{ github.event_name == 'workflow_run' && 'minor' || github.event.inputs.bump_type }}
+        run: |
+          # Docker Hub is the source of truth for versions. Find the highest
```
JoeStech: This extraction of major/minor/hotfix is extremely messy and hard to follow. Let's treat the git tag as the source of truth instead, so that this can all be condensed down into a small git tag -l piped to sed like this:

last_minor="$(
  git tag -l |
    grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' |
    sort -V |
    tail -1 |
    sed -E 's/^v[0-9]+\.([0-9]+)\.[0-9]+$/\1/'
)"

Then the minor version can be incremented and we can tag with the new version.

.github/workflows/build-mcp-image.yml
```diff
@@ -4,7 +4,38 @@ on:
...
+# computing/publishing the same auto-incremented version (alias + git tag).
+# Publishing runs (from main, not dry_run) share a single static group so they
+# queue behind one another. Non-publishing runs (dry_run or non-main) never
+# touch shared version/tag state, so they get a unique per-run group and run
+# concurrently without blocking the publishing queue.
+concurrency:
+  group: >-
+    ${{
+      (
+        ((github.event_name == 'workflow_run' && github.event.workflow_run.head_branch == 'main') || (github.event_name != 'workflow_run' && github.ref == 'refs/heads/main'))
```
JoeStech, line 40: there are a bunch of places where these checks are done, and they're hard to audit this way. Can you please simplify this down to a single "main" branch check at the top of the file, so that this expression isn't repeated over and over

.github/workflows/build-mcp-image.yml
```diff
@@ -22,12 +53,31 @@ jobs:
     runs-on: ${{ matrix.runner }}
     permissions:
       contents: read
-      packages: write
     outputs:
       build_date: ${{ steps.build_date.outputs.build_date }}
     steps:
+      - name: Guard non-main runs against publishing
```
JoeStech: can we just put this check at the top of the file in a less verbose way

.github/workflows/build-mcp-image.yml
```diff
@@ -64,24 +123,406 @@ jobs:
...
+          # Publish only the immutable date tag here. The stable latest/version
+          # aliases are added later, once this merged manifest is verified.
+          # Gated to main so non-main runs never overwrite the shared date tag.
+          if [ "${DRY_RUN}" = "true" ]; then
+            echo "[dry-run] Would create date manifest tag: ${BUILD_DATE}"
+            echo "[dry-run] from sources: ${BUILD_DATE}-amd64, ${BUILD_DATE}-arm64"
+            exit 0
+          fi
+
+          # The date tag is meant to be immutable (it identifies the "stage"
```
JoeStech: this doesn't really work properly, because the actual individual arch images are already tagged with the date.

This is what happens with the same-day issue:

Run A on 2026-06-15:
  builds arch tags
  creates date manifest 2026-06-15
  publishes version 2.3.0 -> 2026-06-15
  creates git tag v2.3.0

Run B later on 2026-06-15:
  builds new arch tags 2026-06-15-amd64 / arm64
  sees date manifest 2026-06-15 already exists
  skips recreating it
  computes version 2.4.0
  aliases 2.4.0 -> existing 2026-06-15 manifest
  creates git tag v2.4.0 on Run B’s commit

So v2.4.0 points at Run B’s commit, but Docker 2.4.0 points at Run A’s image digest.

Let's fix this by just adding a timestamp to the end of the date tag, then we can get rid of all this verbosity.

.github/workflows/build-mcp-image.yml
```diff
@@ -64,24 +123,406 @@ jobs:
...
+          if docker buildx imagetools inspect armlimited/arm-mcp:${BUILD_DATE} >/dev/null 2>&1; then
+            echo "Date tag armlimited/arm-mcp:${BUILD_DATE} already exists; leaving it unchanged (immutable). Skipping manifest creation."
+            exit 0
+          fi
+          docker buildx imagetools create \
+            -t armlimited/arm-mcp:${BUILD_DATE} \
+            armlimited/arm-mcp:${BUILD_DATE}-amd64 \
+            armlimited/arm-mcp:${BUILD_DATE}-arm64
+
+      - name: Verify date-tagged manifest
```
JoeStech: if the build step succeeds this step is not necessary

.github/workflows/build-mcp-image.yml
JoeStech: actually instead of timestamp let's use YYYY-MM-DD-${github.run_number}

.github/workflows/build-mcp-image.yml
```diff
@@ -64,24 +123,406 @@ jobs:
...
+            minor)  TAG_VERSION="${MAX_MAJOR}.$((MAX_MINOR + 1)).0" ;;
+            hotfix) TAG_VERSION="${MAX_MAJOR}.${MAX_MINOR}.$((MAX_PATCH + 1))" ;;
+            *) echo "::error::Unknown bump_type '${BUMP_TYPE}'."; exit 1 ;;
+          esac
+
+          echo "Highest Docker Hub version: ${MAX_MAJOR}.${MAX_MINOR}.${MAX_PATCH}"
+          echo "Next (${BUMP_TYPE}) version tag: ${TAG_VERSION}"
+          echo "tag_version=${TAG_VERSION}" >> "$GITHUB_OUTPUT"
+
+      - name: Add latest and version aliases to verified date tag
```
JoeStech: If we change build to YYYY-MM-DD-${github.run_number} as I mentioned above, we can just drop almost all of this verbose immutability/idempotency code

.github/workflows/build-mcp-image.yml
```diff
@@ -64,24 +123,406 @@ jobs:
...
+            -t armlimited/arm-mcp:latest \
+            -t armlimited/arm-mcp:${TAG_VERSION} \
+            armlimited/arm-mcp:${BUILD_DATE}
+          echo "Aliased ${BUILD_DATE} as latest and ${TAG_VERSION}."
+
+      - name: Create and push matching git tag
+        id: gittag
+        if: ${{ steps.gate.outputs.is_main == 'true' }}
+        env:
+          TAG_VERSION: ${{ steps.version.outputs.tag_version }}
```
JoeStech, line 216: Again, very verbose. If you simplify the workflow so each release computes a fresh tag version, you can make the git tag step fail if the tag already exists. That is easier to reason about.

.github/workflows/build-mcp-image.yml
```diff
@@ -64,24 +123,406 @@ jobs:
...
+            echo "tag_exists=false" >> "$GITHUB_OUTPUT"
+          else
+            git tag -a "${GIT_TAG}" -m "Release ${GIT_TAG} (matches Docker Hub armlimited/arm-mcp:${TAG_VERSION})"
+            git push origin "refs/tags/${GIT_TAG}"
+            echo "Created and pushed git tag ${GIT_TAG}."
+            echo "created=true" >> "$GITHUB_OUTPUT"
+            echo "tag_exists=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Create GitHub Release with changelog
```
JoeStech: This can basically be

- name: Create GitHub Release with changelog
    env:
      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      GIT_TAG: ${{ steps.gittag.outputs.git_tag }}
      TAG_VERSION: ${{ steps.version.outputs.tag_version }}
    run: |
      gh release create "${GIT_TAG}" \
        --title "${GIT_TAG}" \
        --notes "Docker image: armlimited/arm-mcp:${TAG_VERSION}" \
        --generate-notes

.github/workflows/build-mcp-image.yml
```diff
@@ -64,24 +123,406 @@ jobs:
...
+          # Re-read the generated notes and prepend the matching Docker image
+          # reference, then write the combined body back to the release.
+          GENERATED_NOTES="$(gh release view "${GIT_TAG}" --json body --jq '.body')"
+          HEADER="Docker image: \`armlimited/arm-mcp:${TAG_VERSION}\` (built ${BUILD_DATE})"
+          printf '%s\n\n%s\n' "${HEADER}" "${GENERATED_NOTES}" > release_notes.md
+          gh release edit "${GIT_TAG}" --notes-file release_notes.md
+          rm -f release_notes.md
+          echo "Created GitHub Release ${GIT_TAG}."
+
+      - name: Update server.json version
```
JoeStech: This should be done before the build and tagging, because we want the tag to match what's actually in server.json. Right now if we check out the tag, it will get the old version number.

JoeStech: This can also be extremely stripped down, it doesn't need to be over 100 lines.

.github/release.yml
```diff
@@ -0,0 +1,42 @@
...
+      - skip-changelog
+    authors:
+      # Bot accounts author PRs/commits under their "[bot]" login (e.g.
+      # dependabot[bot], github-actions[bot]). The bare names without the
+      # "[bot]" suffix won't match, so exclude both forms to be safe.
+      - dependabot
+      - dependabot[bot]
+      - github-actions
+      - github-actions[bot]
+  categories:
```
JoeStech: We don't use any of these labels, so all of this is unnecessary. Let's remove this file until we need it.

JoeStech (dismissed):

.github/workflows/build-mcp-image.yml
NeethuESim: server.json is updated after the build & tagging is successful, hence it is at the end. If we update it first & then do the build & tag, then in case of failure we will have to revert back the changes. Code after this is handling lot of failure cases which got flagged by co-pilot.

.github/workflows/build-mcp-image.yml
```diff
@@ -55,33 +117,135 @@ jobs:
...
+            echo "[dry-run] Would create ${IMAGE}:${DATE_TAG} from ${DATE_TAG}-amd64/-arm64"
+            exit 0
+          fi
           docker buildx imagetools create \
-            -t armlimited/arm-mcp:${{ env.BUILD_DATE }} \
-            armlimited/arm-mcp:${{ env.BUILD_DATE }}-amd64 \
-            armlimited/arm-mcp:${{ env.BUILD_DATE }}-arm64
+            -t "${IMAGE}:${DATE_TAG}" \
+            "${IMAGE}:${DATE_TAG}-amd64" \
+            "${IMAGE}:${DATE_TAG}-arm64"
```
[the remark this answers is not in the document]
NeethuESim: successful build is enough for verification

.github/workflows/bump-server-json.yml
```diff
@@ -0,0 +1,138 @@
...
+          # inputs.tag is provided by both workflow_call and workflow_dispatch.
+          GIT_TAG: ${{ inputs.tag }}
+        run: |
+          SERVER_JSON="mcp-local/server.json"
+          # Validate the tag as-is (must be a vX.Y.Z release tag) before
+          # deriving the bare X.Y.Z version, so a non-v tag is rejected.
+          if ! printf '%s' "${GIT_TAG}" | grep -Eq '^v[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "::error::Tag '${GIT_TAG}' is not a vX.Y.Z release tag."; exit 1
+          fi
+          TAG_VERSION="${GIT_TAG#v}"
```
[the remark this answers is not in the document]
NeethuESim: workflow_dispatch is removed, server.json is updated only through workflow_call

.github/workflows/bump-server-json.yml
```diff
@@ -0,0 +1,131 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GIT_TAG: ${{ inputs.tag }}
+        run: |
+          SERVER_JSON="mcp-local/server.json"
+          # Validate the tag as-is (must be a vX.Y.Z release tag) before
+          # deriving the bare X.Y.Z version, so a non-v tag is rejected.
+          if ! printf '%s' "${GIT_TAG}" | grep -Eq '^v[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "::error::Tag '${GIT_TAG}' is not a vX.Y.Z release tag."; exit 1
+          fi
+          TAG_VERSION="${GIT_TAG#v}"
```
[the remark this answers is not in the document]
NeethuESim: caller of this workflow is build-mcp-image gh, we can ignore this scenario

brikin01 (commented):
Thanks for all these changes! I just have two points:

1. It might be worth blocking publishing reruns. Rerunning the same workflow can calculate a new semver tag and publish another production version for the same build. Since we probably don’t want full release-repair logic, the simple fix is to fail publishing reruns and ask people to start a fresh workflow run instead.

2. I think we should rearrange a few steps. After determining `TAG_VERSION`, we should commit the `server.json` changes before creating the git tag/GitHub Release, so the release points to the commit with matching metadata. We should also move the production Docker tags to the end, so if something fails midway, we only have a dated staged image to retry or clean up, not a partially promoted production release.

Suggested flow:

1. Build/push arch images: `DATE_TAG-amd64`, `DATE_TAG-arm64`
2. Create the multi-arch staged image: `IMAGE:DATE_TAG`
3. Determine `TAG_VERSION`
4. Bump and commit `server.json`
5. Create the git tag and GitHub Release
6. Promote `IMAGE:DATE_TAG` to `IMAGE:${TAG_VERSION}` and `IMAGE:latest`

NeethuESim:
> Thanks for all these changes! I just have two points:
> 
> 1. It might be worth blocking publishing reruns. Rerunning the same workflow can calculate a new semver tag and publish another production version for the same build. Since we probably don’t want full release-repair logic, the simple fix is to fail publishing reruns and ask people to start a fresh workflow run instead.
> 2. I think we should rearrange a few steps. After determining `TAG_VERSION`, we should commit the `server.json` changes before creating the git tag/GitHub Release, so the release points to the commit with matching metadata. We should also move the production Docker tags to the end, so if something fails midway, we only have a dated staged image to retry or clean up, not a partially promoted production release.
> 
> Suggested flow:
> 
> 1. Build/push arch images: `DATE_TAG-amd64`, `DATE_TAG-arm64`
> 2. Create the multi-arch staged image: `IMAGE:DATE_TAG`
> 3. Determine `TAG_VERSION`
> 4. Bump and commit `server.json`
> 5. Create the git tag and GitHub Release
> 6. Promote `IMAGE:DATE_TAG` to `IMAGE:${TAG_VERSION}` and `IMAGE:latest`

Thanks for the review. 2 is a valid concern. However if steps 5 or 6 fail then we will have to revert server.json & additional clean ups will be needed. Working on re-ordering these steps for correct updation.

NeethuESim:
Current order is as below -

1. Create dated multi-arch manifest
2. Determine version tag → TAG_VERSION
3. Bump server.json on main (commits first, exports bump_sha)
4. Add latest + TAG_VERSION Docker aliases
5. Create git tag + GitHub Release (at bump_sha), last 

Now, git tag is the version cursor, and it's created last, so anything failing before it leaves the cursor unmoved and a re-run recomputes the same version (self-healing). So we need to make sure if this gh action fails at any step, it needs to be re-run to get everything updated.

image

This last scenario here can be ignored, since it is very rare case as window is very tiny. This could also be recovered manually by deleting the orphan tag or creating the missing release. 

@brikin01 please review & let me know if this logic makes sense.

.github/workflows/build-mcp-image.yml
```diff
@@ -55,33 +126,204 @@ jobs:
...
+          apply_bump_file() {
+            local updated
+            if ! updated="$(jq --arg v "${TAG_VERSION}" '
+              .version = $v
+              | .packages |= map(
+                  if (.identifier // "") | startswith("docker.io/armlimited/arm-mcp:")
+                  then .identifier = "docker.io/armlimited/arm-mcp:\($v)"
+                  else . end
+                )
+            ' "${SERVER_JSON}")"; then
```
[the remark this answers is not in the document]
brikin01: Easy and sensible suggestion I think

.github/workflows/build-mcp-image.yml
```diff
@@ -55,33 +126,204 @@ jobs:
...
+        id: version
+        env:
+          BUMP_TYPE: ${{ github.event_name == 'workflow_run' && 'minor' || github.event.inputs.bump_type }}
+        run: |
+          set -euo pipefail
+          git fetch --tags --force --quiet
+          # `|| true`: with pipefail, grep exits non-zero when no tag matches;
+          # the explicit empty-check below emits a clearer error than set -e.
+          latest="$(git tag -l | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -1)" || true
+          [ -n "${latest}" ] || { echo "::error::No vX.Y.Z tags found."; exit 1; }
```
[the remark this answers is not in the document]
brikin01: I think this one could be ignored given we already have an existing tag

.github/workflows/build-mcp-image.yml
```diff
@@ -55,33 +126,204 @@ jobs:
...
-      packages: write
+      contents: write
     if: ${{ needs.build-arch-images.result == 'success' }}
     steps:
+      - name: Block publishing reruns
+        # Repeated from build-arch-images: "Re-run failed jobs" reruns only
+        # this job, so that job's guard won't fire, but publishing happens
+        # here. See that step for the full rationale (reruns republish a new
+        # version). Start a fresh run to recover instead.
+        if: ${{ env.IS_DRY_RUN != 'true' && github.run_attempt != '1' }}
```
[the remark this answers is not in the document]
brikin01: Minor issue but worth fixing

.github/workflows/build-mcp-image.yml
```diff
@@ -55,33 +126,204 @@ jobs:
...
+          if git push origin HEAD:main; then
+            echo "Pushed server.json bump to ${TAG_VERSION}."
+            echo "bump_sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          echo "::error::Push rejected: origin/main moved during the run. Start a fresh workflow run."; exit 1
+
+      # Promote production Docker tags before the git tag (the version cursor)
+      # advances. If this fails, no tag exists yet, so a re-run recomputes the
+      # SAME version; imagetools create overwrites, so re-runs are safe.
```
brikin01: If I'm understanding the code correctly, we block a Github actions re-run and instead it's a fresh workflow run that 'self-heals' (as described in the comment in lines 77-81. Given that, could this comment just be clarified a bit to reference fresh workflow run instead of re-run?

.github/workflows/build-mcp-image.yml
```diff
@@ -55,33 +126,204 @@ jobs:
...
+            -t "${IMAGE}:latest" \
+            -t "${IMAGE}:${TAG_VERSION}" \
+            "${IMAGE}:${DATE_TAG}"
+          echo "Aliased ${DATE_TAG} as latest and ${TAG_VERSION}."
+
+      # FINAL irreversible step: the git tag is the version cursor, so create
+      # it last. Every earlier step (bump, Docker aliases) is idempotent and
+      # leaves no tag on failure, so a re-run recomputes the SAME version and
+      # self-heals. Tag and Release are created together (both at the bump
+      # commit) to minimize the window between them.
```
brikin01: Similar comment as above on 're-run' terminology

brikin01 (changes requested):
Thanks for iterating on this. The updated ordering makes sense to me as a reasonable tradeoff: the git tag remains the version cursor, and failures before that point can be recovered with a fresh workflow run rather than GitHub’s rerun button.

I left a couple of small comments, mostly to clarify the workflow comments/error text so the recovery path is explicit. I also agree with the two minor Copilot comments (ignoring the “no existing version tag” one).

With just those minor changes, happy to approve. Really appreciate all the back and forth on this; there are definitely a bunch of edge cases in this type of workflow.

NeethuESim:
> Thanks for iterating on this. The updated ordering makes sense to me as a reasonable tradeoff: the git tag remains the version cursor, and failures before that point can be recovered with a fresh workflow run rather than GitHub’s rerun button.
> 
> I left a couple of small comments, mostly to clarify the workflow comments/error text so the recovery path is explicit. I also agree with the two minor Copilot comments (ignoring the “no existing version tag” one).
> 
> With just those minor changes, happy to approve. Really appreciate all the back and forth on this; there are definitely a bunch of edge cases in this type of workflow.

Thanks for your review @brikin01, need to get as many eyes on this as possible to not miss any corner case. I have fixed all the open issues.

brikin01 (approved):
LGTM!

PR #94, opened 2026-06-18
Selectively include developer.arm.com content

apickard:
Uses the developer.arm.com search api to retrieve search results, filters them to just the relevant ones, and then generates embeddings from them. Implemented as separate script generate-vectors.py that has the same command line as generate-chunks.py (it needs one argument, the vector csv file). Functions common to both generate-chunks.py and generate-vectors.py have been moved into generate_common.py.


apickard:
Will be replaced by another PR that includes fixed unit tests.

PR #97, opened 2026-06-23
Updated vector-db-source csv with education content

tomcarter868:
Sixth column added for education content.


NeethuESim:
PR comments will be incorporated in [STESOL-526](https://jira.arm.com/browse/STESOL-526)

PR #109, opened 2026-07-07
[STESOL-554] Have eval script use main search function and add sources to eval answers

brikin01:
## Summary
- Route retrieval evaluation through the shared `arm_kb_search.search()` path so eval matches production search behavior
- Set explicit search candidate depth from requested result count, avoiding accidental coupling to deduplication candidate count
- Add alternate expected URLs for eval questions where multiple relevant sources satisfy the query.
- Bump package version to `0.2.0`.

### DIFF
--- a/arm_kb_search/resources.py
+++ b/arm_kb_search/resources.py
@@ -113,6 +113,7 @@ def search(
     k: int | None = None,
 ) -> list[dict[str, Any]]:
     resolved_k = k or resources.default_k
+    candidate_depth = max(resolved_k * 20, 100)
     search_results = hybrid_search(
         query,
         resources.usearch_index,
@@ -120,6 +121,7 @@ def search(
         resources.embedding_model,
         resources.bm25_index,
         k=deduplication_candidate_count(resolved_k),
+        candidate_depth=candidate_depth,
     )
     deduped = deduplicate_urls(search_results)[:resolved_k]
     formatted = [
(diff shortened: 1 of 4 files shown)


arm_kb_search/resources.py
```diff
@@ -113,13 +113,15 @@ def search(
     k: int | None = None,
 ) -> list[dict[str, Any]]:
     resolved_k = k or resources.default_k
+    candidate_depth = max(resolved_k * 20, 100)
```
NeethuESim, line 116: @brikin01 why is this line added here ? Won't this reduce the number of relevant chunks ?

brikin01: This line makes the default runtime search use the same candidate-depth calculation as the original `evaluate_retrieval.py` (see L59 of the original file).

For a requested `k=5`, that means examining 100 dense and BM25 candidates. Without this line, `hybrid_search()` calculates its default from `deduplication_candidate_count(k)`, which is 50, resulting in a candidate depth of 1,000.

Reducing the depth from 1,000 to 100 can theoretically exclude a relevant chunk ranked below 100 in the initial retrieval stage, but it does not reduce the requested number of final results (although as mentioned above, we were never actually using a depth of 1000 in this particular eval script). Also, I tested the retrieval evaluation with both values. 100 produced equivalent or slightly better results while requiring less search and reranking work, so 100 seems like the better tradeoff here.

NeethuESim (approved):
LGTM

PR #110, opened 2026-07-07
Add Arm enablement workflow and Codex skill

ranimandepudi:
## Summary

Adds a reusable Arm enablement workflow for OSS repositories in VS Code and Codex.

The Codex integration is packaged as an `arm-enablement` skill. It uses the Arm MCP Server to assess source, builds, dependencies, CI, and container supply chains, then creates evidence-backed `arm-enablement-report.md` and `arm-enablement-report.pdf` artifacts. Report-only mode is the default; fixes and commits require explicit user requests.

## What changed

- Add the VS Code chat integration for Arm readiness assessments.
- Add the Codex `arm-enablement` skill with `agents/openai.yaml` interface metadata.
- Keep Arm MCP setup manual because the skill dependency schema cannot express the required Docker image, arguments, and workspace mount.
- Document Codex skill installation, discovery with `/skills`, invocation with `$arm-enablement`, and the `/workspace` mount requirement.
- Handle unsupported scanner languages explicitly instead of substituting an unrelated scanner.
- Scope knowledge-base searches to decisions that affect the verdict and treat an empty result as no evidence.
- Use local `/workspace` scans for changed checkouts and reserve `git_repo` for an initial remote report-only scan.
- Inspect digest pins as manifest lists or single-platform images instead of assuming multi-architecture support.
- Keep report-only runs limited to the Markdown and PDF artifacts, with no automatic commits.
- Record MCP arguments, purpose, result, and measured client-observed duration in readable three-column report tables.

## Validation

- The Codex skill passes the official skill validator, YAML parsing, frontmatter, tool-list, and whitespace checks.
- A fresh end-to-end Codex 0.144.2 run loaded `$arm-enablement` from `$HOME/.agents/skills`, connected through the documented manual Arm MCP configuration, and assessed pinned `etcd-io/bbolt` commit `fa7ad3a72579c8766a84f7e765a4a9de997e0cf6` in report-only mode.
- The run generated only `arm-enablement-report.md` and `arm-enablement-report.pdf`. The final PDF is 11 pages; every page was rendered and checked for clipping, overlap, blank pages, duplicate titles, table readability, and pagination.
- On a native AWS `aarch64` Ubuntu host, `armlimited/arm-mcp:latest` resolved to Arm64 digest `sha256:d40b728c7f060915879bbd64a7ff25bf2ba77e44876f19775c21e8d25b681b76`. Arm MCP Server 3.4.4 exposed all seven expected tools, and the Go scan completed with return code 0 and zero findings.
- bbolt built and executed with Go 1.25.12 on native `linux/arm64`. Its documented short test matrix and short race matrix passed for both freelist implementations. The temporary AWS checkout was clean after removing the MCP invocation-log sidecar.

## CI

- `integration-tests (ubuntu-24.04, linux/amd64)` passed on commit `3704576`.
- `integration-tests (ubuntu-24.04-arm, linux/arm64)` passed on commit `3704576`.


agent-integrations/codex/arm-enablement.md
```diff
@@ -0,0 +1,107 @@
+<!-- Place this prompt file at ~/.codex/prompts/arm-enablement.md to enable it.
+     Invoke using /prompts:arm-enablement in the codex chat.
+-->
+---
+description: Scan an OSS codebase with Arm MCP and generate a CubeFS-style Arm enablement report as Markdown and PDF
```
brikin01: Do we want to mention `CubeFS-style`? Reads a bit specific to me, will users know what that refers to?

agent-integrations/codex/arm-enablement.md
```diff
@@ -0,0 +1,107 @@
...
+Input handling:
+
+* If the user runs `/prompts:arm-enablement` from an already-cloned OSS project, analyze the current workspace.
+* If the user provides a GitHub repository URL, clone that repository into the current workspace if it is empty, or into a clearly named subdirectory such as `arm-enablement-target/` if the workspace is not empty. Then analyze the cloned repository.
+* If the user provides `--apply-fixes` or explicitly asks to fix the project, apply the minimum source changes needed for Arm enablement after writing the initial report, then update the report with the implemented fixes and re-run the relevant MCP checks.
+* If the user does not request fixes, stay in report-only mode. Do not modify project source files except for creating `arm-enablement-report.md` and `arm-enablement-report.pdf`.
+
+Steps to follow:
+
+* Detect the project's primary language(s) by inspecting the codebase (`go.mod`, `package.json`, `requirements.txt`, `pom.xml`, `Cargo.toml`, `CMakeLists.txt`, source file extensions). Pick the appropriate `migrate_ease_scan` scanner: `cpp`, `python`, `go`, `js`, or `java`.
```
brikin01: This line mentions `Cargo.toml`, which is for Rust, but `migrate_ease_scan` doesn't have a Rust scanner. Should we include instructions on what to do when there's an unsupported language?

agent-integrations/codex/arm-enablement.md
```diff
@@ -0,0 +1,107 @@
+<!-- Place this prompt file at ~/.codex/prompts/arm-enablement.md to enable it.
```
brikin01: [Custom prompt files are now deprecated for Codex](https://learn.chatgpt.com/docs/custom-prompts) and [are not invokable in more recent versions](https://github.com/openai/codex/issues/15972) of Codex CLI. Can we convert this into a skill file?

agent-integrations/codex/arm-enablement.md
```diff
@@ -0,0 +1,107 @@
...
+     Invoke using /prompts:arm-enablement in the codex chat.
+-->
+---
+description: Scan an OSS codebase with Arm MCP and generate a CubeFS-style Arm enablement report as Markdown and PDF
+---
+
+Before starting, verify that the `arm-mcp` MCP server is installed and available. If you don't have access to the arm-mcp tools (skopeo, check_image, knowledge_base_search, migrate_ease_scan, mca, apx_recipe_run, sysreport_instructions), refer to the [MCP Server Installation Guide](https://github.com/arm/mcp/blob/main/agent-integrations/agent-install-instructions.md) to install it on codex.
+
+Your goal is to evaluate an open-source codebase for Arm (aarch64) readiness and generate a polished **Arm Enablement Report** in the style of the CubeFS case study, "Arm MCP Server in Action: Enabling multi-arch support for CubeFS." The report must read like a professional external case study for an OSS maintainer, developer-relations team, or CNCF community audience, not like an internal checklist. It must answer: "What is needed to make this project 100% Arm-enabled, and what did the Arm MCP Server discover that ordinary manual review could miss?"
```
brikin01: Is this CubeFS case study easily findable? A quick Google search doesn't reveal a quick hit. If we reference this, maybe we include a link where we know it will be found?

agent-integrations/codex/arm-enablement.md
```diff
@@ -0,0 +1,107 @@
...
+* For each Dockerfile, Compose file, and Kubernetes manifest in the repo, list every container image referenced. For each image, call `arm-mcp/check_image` to confirm `linux/arm64` is published. For images pinned by `@sha256:` digest, also call `arm-mcp/skopeo` with `raw=true` to confirm whether the digest resolves to a multi-arch manifest list or a single-arch manifest. Flag any image that is amd64-only or pinned to a single-arch digest.
+* For each dependency declared in package manifests (Dockerfile `apt-get`/`yum`/`apk` lines, `requirements.txt`, `go.mod`, `package.json`, `pom.xml`), call `arm-mcp/knowledge_base_search` and explicitly ask "Is [package] compatible with Arm architecture?" where [package] is the name of the package. Record the verdict and the recommended version if a change is needed.
```
brikin01: There may not be a satisfactory answer in the knowledge base for many packages, should we mention how to handle that or what should be put into the report? Also how many dependencies do these projects often have? This might be doing a lot of searches and eat up a lot of context.

agent-integrations/codex/arm-enablement.md
```diff
@@ -0,0 +1,107 @@
...
+* If the user provides `--apply-fixes` or explicitly asks to fix the project, apply the minimum source changes needed for Arm enablement after writing the initial report, then update the report with the implemented fixes and re-run the relevant MCP checks.
+* If the user does not request fixes, stay in report-only mode. Do not modify project source files except for creating `arm-enablement-report.md` and `arm-enablement-report.pdf`.
+
+Steps to follow:
+
+* Detect the project's primary language(s) by inspecting the codebase (`go.mod`, `package.json`, `requirements.txt`, `pom.xml`, `Cargo.toml`, `CMakeLists.txt`, source file extensions). Pick the appropriate `migrate_ease_scan` scanner: `cpp`, `python`, `go`, `js`, or `java`.
+* Run `arm-mcp/migrate_ease_scan` against the workspace with the chosen scanner. If the user supplied a GitHub URL and the workspace is not the target checkout, pass the URL as `git_repo` only for the scanner call, then use the local clone for file inspection. Capture every architecture-sensitive finding (file path, line number, category, suggestion). This is the discovery phase and drives the rest of the report.
```
brikin01: Could we clarify that `git_repo` is only for initial report-only scans when the current workspace is unrelated? For uncommitted changes and all `--apply-fixes` validation, require the target repo at `/workspace` and omit `git_repo`; otherwise the scanner might create a fresh clone that excludes local fixes and can produce false validation results.

I guess this somewhat overlaps with line 22 above, maybe we can somehow consolidate or clarify better?

agent-integrations/codex/arm-enablement.md
```diff
@@ -0,0 +1,107 @@
...
+* OPTIONAL: If the user is on an Arm host or has access to an Arm runner (AWS Graviton, Azure Cobalt, GCP Axion), validate the final state with native builds, `file <binary>` architecture checks, `sysreport_instructions`, and `arm-mcp/apx_recipe_run` for performance/hotspot evidence when relevant. If no Arm host is available, mark validation as deferred.
+* Maintain an audit trail: for every `arm-mcp` tool call, record `{timestamp, tool, arguments, reason}`. This drives the "Audit Trail" section of the final report.
+
+Pitfalls to avoid:
+
+* Do not equate `grep -r "amd64"` results with actionable findings. `migrate_ease_scan` filters out false positives in vendored dependencies, test fixtures, and assembly files that already carry arm64 build tags. Trust the scanner over manual grep.
+* Do not assume an image is multi-arch because the upstream tag has Arm64 manifests. A `@sha256:` digest pin in a Dockerfile resolves to a single platform manifest and will fail with `exec format error` on Arm hosts. Always inspect digests with `skopeo`.
```
brikin01: > A `@sha256:` digest pin in a Dockerfile resolves to a single platform manifest

Is this true? I think maybe it can point to an image index which is multi-platform?

agent-integrations/codex/arm-enablement.md
```diff
@@ -0,0 +1,107 @@
...
+After the markdown report is complete, export it to a local PDF named `arm-enablement-report.pdf` at the repo root.
+
+PDF export instructions:
+
+* First try `pandoc arm-enablement-report.md -o arm-enablement-report.pdf` if `pandoc` is installed.
+* If `pandoc` is not available and Node.js is installed, try `npx --yes md-to-pdf arm-enablement-report.md --output arm-enablement-report.pdf`.
+* If both exporters are unavailable, do not invent a PDF. Leave `arm-enablement-report.md` complete and tell the user exactly which command to run after installing `pandoc` or Node.js. The markdown report remains the source of truth.
+* After export, verify that `arm-enablement-report.pdf` exists and is non-empty. If possible, report its file size.
+
+If the user has explicitly asked for changes to be applied, after the report is written: apply the fixes from the Implementation Plan in the workspace, commit each fix as a separate change for review-ability, and re-run `migrate_ease_scan` to confirm findings are resolved. If the user asked only for analysis, do not modify any source files; the report alone is the deliverable.
```
brikin01: I understand applying changes, but maybe it shouldn't commit the fixes? That part should probably be left up to the user--thoughts?

brikin01 (commented):
Thanks for putting this together—I left a few comments, though some are minor, so please use your best judgment. I’m not an expert in this type of analysis, so happy to be corrected if I’ve misunderstood anything

agent-integrations/codex/arm-enablement.md
ranimandepudi: Good Point !! I used it while testing and forgot to remove before the PR push. I removed the CubeFS-style wording and made the guidance project-neutral. The skill now asks for a professional, evidence-backed case-study report for OSS maintainers and community audiences.

agent-integrations/codex/arm-enablement.md
ranimandepudi: Agree and thanks for the catch. I added explicit handling for unsupported languages. The skill now lists the scanners supported by migrate_ease_scan. For an unsupported primary language, it must not substitute another scanner or claim that the scan passed. It records the limitation, manually reviews architecture-sensitive areas, uses cross-build or native Arm evidence, and marks anything not validated as deferred.

agent-integrations/codex/arm-enablement.md
ranimandepudi: Agreed and I converted the legacy codex custom prompt into an arm-enablement skill with SKILL.md and agents/openai.yaml, including the arm mcp dependency. I also added installation and invocation guidance. I tested the skill with Codex 0.144.2 against bbolt and generated both report and artifacts successfully.

agent-integrations/codex/arm-enablement.md
ranimandepudi: Yes and since the CubeFS case study is not a stable public reference, I removed the reference instead of adding a link. The report requirements and structure are now fully self-contained in the skill.

agent-integrations/codex/arm-enablement.md
ranimandepudi: You are right. A digest can identify either a multi-platform image index/manifest list or a single-platform manifest. I corrected the guidance so the skill uses skopeo raw manifest inspection to determine which one it is instead of assuming that a digest is single-platform.

agent-integrations/codex/arm-enablement.md
ranimandepudi: Agreed. Report-only mode is now the default. Fixes are applied only when the user explicitly requests them, and the skill does not commit anything unless the user separately asks for a commit. It also reruns the affected tests and scanner against the changed /workspace checkout before reporting a finding as resolved.

agent-integrations/codex/arm-enablement.md
ranimandepudi: Good point. I updated the skill to avoid searching every dependency. It now focuses on direct build/runtime and architecture-sensitive dependencies, groups related packages, and only queries the knowledge base when the result could affect the verdict or remediation. If no relevant result is returned, the report records No Arm-specific KB result, does not infer incompatibility, and uses upstream documentation, published artifacts, or native validation instead. If no stronger evidence is available, the item is marked unverified. This makes the no-result behavior explicit while keeping context usage bounded.

agent-integrations/codex/arm-enablement.md
ranimandepudi: Agreed. I clarified that git_repo is only for an initial report-only scan when the current workspace is unrelated and there are no local changes to assess. Uncommitted changes and all fix validation must use the changed checkout mounted at /workspace, with git_repo omitted. The skill also verifies that Codex and the MCP container are looking at the same checkout before scanning.

agent-integrations/codex/arm-enablement/agents/openai.yaml
```diff
@@ -0,0 +1,12 @@
...
+  short_description: "Assess OSS projects for Arm readiness"
+  default_prompt: "Use $arm-enablement to assess this repository for Arm readiness and generate the Markdown and PDF report."
+
+dependencies:
+  tools:
+    - type: "mcp"
+      value: "arm-mcp"
+      description: "Arm MCP Server for Arm readiness assessment and validation"
+      transport: "stdio"
+      command: "docker"
```
brikin01: I think we should probably remove this dependencies block from openai.yaml. To my understanding, Codex cannot include the required Docker arguments here, so it would install command = "docker" with no arguments and fail to start Arm MCP. Keep the existing manual MCP configuration and retain only the interface metadata.

agent-integrations/vs-code/arm-enablement.prompt.md
```diff
@@ -0,0 +1,114 @@
...
+* If the codebase contains assembly (`.s`, `.S`) or architecture-specific intrinsics (SSE/AVX, NEON), use `arm-mcp/mca` to analyze representative hot paths and use `arm-mcp/knowledge_base_search` to find the Arm equivalent (NEON, SVE, or SVE2 depending on target).
+* OPTIONAL: If the user is on an Arm host or has access to an Arm runner (AWS Graviton, Azure Cobalt, GCP Axion), validate the final state with native builds, `file <binary>` architecture checks, `sysreport_instructions`, and `arm-mcp/apx_recipe_run` for performance/hotspot evidence when relevant. If no Arm host is available, mark validation as deferred.
+* Maintain an audit trail: for every `arm-mcp` tool call, record `{timestamp, tool, arguments, reason}`. This drives the "Audit Trail" section of the final report.
```
brikin01: The Codex skill also mentions duration, which could be worth adding here too

agent-integrations/vs-code/arm-enablement.prompt.md
```diff
@@ -0,0 +1,114 @@
...
+  * `Step 5: Implementation Plan or Implemented Fixes` with a prioritized table. If report-only, describe the exact maintainer PR that should be opened.
+  * `Step 6: Validation` with build/test status and `file <binary>` architecture output when available.
+* **The Critical Discovery / The Key Discovery** — If a non-obvious bug exists, explain the bug, why it survived, user impact, and one-line fix. If no critical bug exists, use `The Key Discovery: What Arm MCP Proved` and describe the risks ruled out plus the strongest remaining parity gap.
+* **The Cost of Leaving It Unchecked** — Explain likely project/user/ecosystem cost. Keep this evidence-based; do not invent adoption numbers.
+* **Effort Comparison** — Three-row table: `Manual Investigation`, `AI agent without Arm MCP`, `AI agent with Arm MCP`; include risk and time-to-discovery.
+* **What Did Not Have to Happen** — Two-column table comparing the traditional approach with the Arm MCP-assisted outcome.
+* **Audit Trail** — Numbered table of every `arm-mcp` tool invocation: `#`, `Time (UTC)`, `Tool`, `Purpose`. End with total invocation count and total MCP computation time.
```
brikin01: Similar to above comment, Codex skill has more specific language on duration

brikin01 (commented):
Thanks for all the changes, looking really good! I just added a few more comments, most important one is about the MCP dependency block (I think we can just delete). The others (along with the comments Copilot generated) are fairly minor, feel free to treat those as non-blocking if you think they're not necessary.

Also don't worry about the failing integration test, it is unrelated to your change and I am fixing that in a separate PR.

ranimandepudi:
Addressed all the comments and made a fresh commit accordingly !! Thanks

brikin01 (approved):
LGTM thanks for all the changes and for adding this into the repo!

PR #107, opened 2026-07-03
Update eval_questions.json

dav-mac:
Added eval questions for education content. One question per file added, including ppt.

PR #112, opened 2026-07-08
feat: STESOL 526/ 542 - Support video transcript/Jupyter Notebook/PPTX chunking in the Arm Knowledge Base

NeethuESim:
### PR Description

Support video transcript/Jupyter Notebook/PPTX chunking in the Arm Knowledge Base

Jira Ticket - 

- https://jira.arm.com/browse/STESOL-526
- https://jira.arm.com/browse/STESOL-542

_Note: Based on latest vector-db-sources.csv with Tom's changes (https://github.com/arm/mcp/pull/97) + eval_questions.json from David (https://github.com/arm/mcp/pull/107) MRR on running `run-question-eval.sh` is 0.901 with 30 Misses. Accuracy difference to be investigated as part of - https://jira.arm.com/browse/STESOL-560._

image

### Testing Steps

- All test cases should pass
- ./run-question-eval.sh should work without any errors. Steps to run are here - https://github.com/arm/mcp/tree/main/embedding-generation


brikin01:
Don't we need to add back in the sources removed previously in #98? They are listed in [STESOL-542](https://jira.arm.com/browse/STESOL-542).

NeethuESim:
> Don't we need to add back in the sources removed previously in #98? They are listed in [STESOL-542](https://jira.arm.com/browse/STESOL-542).

Yup, need to do that. I have been testing with those. Thanks for the reminder, will check them in this PR.

embedding-generation/document_chunking.py
```diff
@@ -497,6 +509,286 @@ def parse_pdf(pdf_bytes: bytes, source_url: str, resolved_url: str, fallback_tit
...
+    parsed_document = parse_markdown(
+        notebook_to_markdown(notebook_data, fallback_title),
+        source_url,
+        resolved_url,
+        fallback_title,
+    )
+    parsed_document.content_type = "notebook"
+    return parsed_document
+
+
```
brikin01, line 789: Have we considered using [nbconvert](https://nbconvert.readthedocs.io/en/latest/index.html) for notebook parsing and Markdown conversion? It appears to cover much of the functionality implemented here, which could reduce the custom conversion code and allow some low-level tests to be consolidated into a smaller set of integration tests. It is also a mature, actively maintained Jupyter project.

embedding-generation/README.md
```diff
@@ -48,15 +48,34 @@ python discover-developer-arm-com-sources.py vector-db-sources.csv
...
+at their primary `URL`. For these, populate the optional `Transcript Source URL`
+column with a link to a plain-text, markdown, PowerPoint (`.pptx`), or Jupyter
+notebook transcript (such as a GitHub `.../blob/...` file). When
+`Transcript Source URL` is set,
+`generate-chunks.py` fetches and chunks the transcript instead of the primary
+`URL`, but keeps the primary `URL` as the user-facing link returned by retrieval:
+
+```csv
+Site Name,License Type,Display Name,URL,Keywords,Transcript Source URL
+Educational Course,All rights reserved,Example Video,https://courses.edx.org/videos/...,arm, ai, inference,https://github.com/arm-education/.../M1KV1.txt
```
brikin01: Nit, but the keywords here should be separated by semicolons, not commas, so as not to have a malformed CSV

embedding-generation/document_chunking.py
```diff
@@ -497,6 +509,286 @@ def parse_pdf(pdf_bytes: bytes, source_url: str, resolved_url: str, fallback_tit
...
+                if document_title == fallback_title:
+                    first_line = next((block.text for block in blocks if block.text), "")
+                    if first_line and len(first_line.split()) <= 12:
+                        document_title = first_line
+                sections.append(Section([f"Slide {slide_number}"], blocks))
+    except BadZipFile:
+        sections = []
+
+    if not sections:
+        sections.append(Section([], [Block("paragraph", fallback_title)]))
```
brikin01: Could we return no sections here, similar to how empty Arm API content is handled? If the URL returns a broken PPTX, Git LFS pointer, or a 200 error page, we currently treat it as successfully ingested and index only the fallback title. It would be better to log it, generate no chunks, and continue with the remaining sources. The invalid-PPTX test could then check that no chunks are generated.

embedding-generation/tests/test_generate_chunks.py
```diff
@@ -39,6 +42,85 @@ def _arm_api_response(title, html):
...
+                    'Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/notesSlide" '
+                    f'Target="../notesSlides/notesSlide{slide_number}.xml"/>'
+                    "</Relationships>"
+                ),
+            )
+            archive.writestr(f"ppt/notesSlides/notesSlide{slide_number}.xml", _pptx_slide_xml(notes[index]))
+
+    return buffer.getvalue()
+
+
```
brikin01: Would it be simpler to use a small checked-in PPTX fixture here instead of building the OOXML package by hand? A real two-slide deck with reordered slides and speaker notes could cover most of these cases, remove quite a bit of test setup, and give us more confidence that this works with actual PowerPoint files.

embedding-generation/tests/test_generate_chunks.py
```diff
@@ -870,6 +1264,199 @@ def fake_fetch(url):
...
+            transcript_url=transcript_url,
+        )
+
+        assert fetched_urls == [raw_transcript_url]
+        assert len(chunks) == 1
+        assert chunks[0].url == source_url
+        assert chunks[0].resolved_url == raw_transcript_url
+        assert chunks[0].content_type == "pptx"
+        assert "Generative AI inference on Arm CPUs" in chunks[0].content
+        assert "<a:t>" not in chunks[0].content
```
brikin01: Are the PPTX and notebook transcript tests needed here? In the CSV these files are used as the primary URL, while the transcript column is blank. It might be simpler to test them through the normal source parsing path and keep the transcript tests focused on sources that actually have a separate transcript.

embedding-generation/tests/test_generate_chunks.py
NeethuESim: Its definitely simpler, tradeoff is the additional pptx file in code. This method adds more lines of code but additional file is added. Am good either way.

embedding-generation/document_chunking.py
NeethuESim: Good suggestion :+1:, will refactor after investigating the accuracy issue, which I will cover in this PR itself.

embedding-generation/tests/test_generate_chunks.py
NeethuESim: Good catch, removed them.

embedding-generation/tests/test_generate_chunks.py
brikin01: Since `embedding-generation` isn’t copied into the MCP image, I don’t think a small PPTX fixture would have any meaningful impact. I’d slightly prefer the fixture because it makes the tests and expected chunking behavior easier to understand and gives us coverage with a real PowerPoint file. That said, this is minor and not a blocker.

NeethuESim: Done.

embedding-generation/vector-db-sources.csv
```diff
@@ -1840,54 +1841,78 @@ Learning Paths,CC4.0,Learning Path - Deploy multi-network device meshes using De
...
+Learning Paths,CC4.0,Learning Path - Deploy an Amazon EKS cluster with AWS Graviton-based nodes using Rafay,https://learn.arm.com/learning-paths/servers-and-cloud-computing/rafay-eks/,Containers and Virtualization; AWS; Linux; Kubernetes; AWS Elastic Kubernetes Service (EKS); Rafay; NGINX; rctl; AI-assisted,
+Learning Paths,CC4.0,Learning Path - Measure and modify Go garbage collection behavior on AWS Graviton-based compute,https://learn.arm.com/learning-paths/servers-and-cloud-computing/go-gc-default-settings/,Performance and Architecture; AWS; Linux; Go; Benchstat,
+Learning Paths,CC4.0,Learning Path - Deploy a machine learning application to the Arm Ethos-U65 NPU on NXP FRDM i.MX 93 with Topo,https://learn.arm.com/learning-paths/embedded-and-microcontrollers/deploy-ml-model-to-npu-with-topo/,Containers and Virtualization; Linux; macOS; Windows; Topo; Docker; SSH; ExecuTorch; remoteproc-runtime,
+Learning Paths,CC4.0,Learning Path - Create an interactive shell for Zephyr RTOS on Arm Cortex-M,https://learn.arm.com/learning-paths/embedded-and-microcontrollers/zephyr_shell/,RTOS Fundamentals; RTOS; Zephyr; C,
```
brikin01, line 1849: Just curious, where did these new learning paths come from?

embedding-generation/vector-db-sources.csv
```diff
@@ -1840,54 +1841,78 @@ Learning Paths,CC4.0,Learning Path - Deploy multi-network device meshes using De
...
+Educational Course,All rights reserved,"Introduction to Machine Learning on constrained devices, Part 1",https://courses.edx.org/videos/block-v1:ArmEducationX+EDARMXML.6x+2T2022+type@video+block@1b743b1f9962476b8f683ad04aa816a5,Machine Learning,https://github.com/arm-education/education-content-for-mcp/blob/main/machine_learning_at_the_edge/machine_learning_at_the_edge/Module%202/M2_SV1.txt
+Educational Course,All rights reserved,"Introduction to Machine Learning on constrained devices, Part 2",https://courses.edx.org/videos/block-v1:ArmEducationX+EDARMXML.6x+2T2022+type@video+block@f1488dc7a38a467b9010c9363e180692,Machine Learning,https://github.com/arm-education/education-content-for-mcp/blob/main/machine_learning_at_the_edge/machine_learning_at_the_edge/Module%202/M2_SV2.txt
+Educational Course,All rights reserved,"Artificial neural networks, Part 1",https://courses.edx.org/videos/block-v1:ArmEducationX+EDARMXML.6x+2T2022+type@video+block@b753760a45304583842fe82d1cc4f1fb,Artificial Neural Networks,https://github.com/arm-education/education-content-for-mcp/blob/main/machine_learning_at_the_edge/machine_learning_at_the_edge/Module%203/M3_SV1.txt
```
brikin01, line 1890: A lot of these video links seem to have lost meaningful keywords as part of this change, is that by design?

embedding-generation/vector-db-sources.csv
```diff
@@ -1840,54 +1841,78 @@ Learning Paths,CC4.0,Learning Path - Deploy multi-network device meshes using De
...
+Educational Resource,Arm Education EULA,Get Practical with ExecuTorch on Arm: Accelerating ExecuTorch on Arm Ethos-U Neural Processors,https://github.com/arm-education/executorch_on_arm_labs/blob/main/Lab_3_Accelerating_ExecuTorch_Ethos_NPU.ipynb,ExecuTorch,
+Educational Resource,Arm Education EULA,Introduction to Cloud Compute on Arm: Introduction to Neoverse Cores,https://github.com/arm-education/Introduction-to-Cloud-Compute-on-Arm/blob/main/Part%201%20-%20Introduction%20to%20Neoverse%20Cores.md,Neoverse; Neoverse cores; Neoverse N1; Neoverse N2; Neoverse V1; Neoverse V2; cloud instance selection; performance; power efficiency; scalability; security; server workload,
+Educational Resource,Arm Education EULA,Introduction to Cloud Compute on Arm: The Software Stack,https://github.com/arm-education/Introduction-to-Cloud-Compute-on-Arm/blob/main/Part%202%20-%20The%20Software%20Stack.md,AI,
+Educational Resource,Arm Education EULA,Introduction to Cloud Compute on Arm: Arm Neoverse Software and System Design,https://github.com/arm-education/Introduction-to-Cloud-Compute-on-Arm/blob/main/Part%203%20-%20Arm%20Neoverse%20Software%20and%20System%20Design.md,AI,
```
brikin01, line 1915: A few of these are quite detailed on the keywords, but others have only one and oftentimes that one is present in the title. Was that on purpose?

embedding-generation/document_chunking.py
brikin01: Were you still planning on refactoring this as part of this PR? Could open a prospective Jira ticket for follow-up work if it's not to be prioritized now.

embedding-generation/eval_questions.json
```diff
@@ -1,4 +1,405 @@
...
+    "https://courses.edx.org/videos/block-v1:ArmEducationX+EDARMXML.6x+2T2022+type@video+block@5eee196dbadb407cbbc998d96ebed206",
+    "https://learn.arm.com/learning-paths/embedded-and-microcontrollers/img_nn_stcube"
+  ]
+  },
+  {
+  "question": "How can I quantize a TensorFlow model with TensorFlow Lite and verify the improvements in memory usage, inference latency, and energy consumption on an STM32 microcontroller using STM32Cube.AI?",
+  "expected_urls": [
+    "https://courses.edx.org/videos/block-v1:ArmEducationX+EDARMXML.6x+2T2022+type@video+block@22b73d7ee4e94d738ae8dcf58697c7c6",
+    "https://courses.edx.org/videos/block-v1:ArmEducationX+EDARMXML.6x+2T2022+type@video+block@a1297eceb2c341e0b6b449ee625843d4"
+  ]
```
brikin01, line 276: Some of the newly added acceptable URLs appear to mask retrieval failures rather than represent equally good answers. For this question, the first URL directly covers TensorFlow Lite quantization, STM32Cube.AI, memory, latency, and energy. The second discusses general Cortex-M resource constraints but does not cover TensorFlow Lite, STM32Cube.AI, or quantization.

Maybe we keep only the first URL as expected? Did you notice other examples like this? These might be useful “hard” cases for improving retrieval rather than broadening the acceptable answers. Let me know what you think

Given this eval set keeps growing, it might be worth a conversation amongst ourselves (and Joe when he gets back) as to what we're aiming for with these evaluation questions and how we can best add questions without also needed to debug accuracy each time.

embedding-generation/document_chunking.py
NeethuESim: I am planning to look into this, but will create a separate jira item. This PR is already too big now.

embedding-generation/vector-db-sources.csv
NeethuESim: So I have been trying to balance out the keywords to handle the ranking. Earlier all keywords were mostly similar for all this new content. Some, like in other examples needed more specific keywords to get ranked higher. Goal was to get the questions passed. Definitely not a good approach, since as new sources get added, ranking will change & this might affect the final result. I agree this might be case to also improve the retrieval algorithm. We can discuss when Joe is back.

embedding-generation/eval_questions.json
NeethuESim: I am looking into this one, why the other link was added. Maybe was a mistake.

embedding-generation/vector-db-sources.csv
NeethuESim: If i remember correctly, there was some difference in Tom's list & what I had earlier. So got added in that commit. But now I look in his changes, i don't see it. So let me double check, if not then will remove this.

NeethuESim: Looks like its getting added after running the run_evaluation scripts, probably during the discovery phase -

[NEW SOURCE] Learning Path - Deploy Gerrit on a Google Cloud C4A instance: https://learn.arm.com/learning-paths/servers-and-cloud-computing/gerrit-on-gcp/
[NEW SOURCE] Learning Path - Deploy containers on Arm-based compute using Amazon ECS and the AWS CDK: https://learn.arm.com/learning-paths/servers-and-cloud-computing/aws-cdk/
[NEW SOURCE] Learning Path - Deploy an Amazon EKS cluster with AWS Graviton-based nodes using Rafay: https://learn.arm.com/learning-paths/servers-and-cloud-computing/rafay-eks/
[NEW SOURCE] Learning Path - Measure and modify Go garbage collection behavior on AWS Graviton-based compute: https://learn.arm.com/learning-paths/servers-and-cloud-computing/go-gc-default-settings/
[NEW SOURCE] Learning Path - Deploy a machine learning application to the Arm Ethos-U65 NPU on NXP FRDM i.MX 93 with Topo: https://learn.arm.com/learning-paths/embedded-and-microcontrollers/deploy-ml-model-to-npu-with-topo/
[NEW SOURCE] Learning Path - Create an interactive shell for Zephyr RTOS on Arm Cortex-M: https://learn.arm.com/learning-paths/embedded-and-microcontrollers/zephyr_shell/

embedding-generation/eval_questions.json
NeethuESim: As you mentioned this is one of those "hard" cases for improving retrieval. Weird thing is even after I added the keywords specific to questions in this source link, it still is not getting ranked on the top. https://github.com/NeethuESim/mcp/blob/4b3a76e2a8e4e7cdf247a7a59af0621bb782ab92/embedding-generation/vector-db-sources.csv?plain=1#L1894
I can remove the 2nd link & open another jira ticket to track this behavior. It currently does fails as a miss when only the 1st link is the expected value.

embedding-generation/document_chunking.py
NeethuESim: Jira created - https://jira.arm.com/browse/STESOL-574

embedding-generation/eval_questions.json
brikin01: Yeah it is odd and definitely an area where our retrieval needs to improve. I think better to have a few eval questions that fail instead vs including a link that's only tangentially related. If there were any others that were on the border, definitely flag those too, either in the Jira or by removing a tangential second link in the eval.

embedding-generation/vector-db-sources.csv
brikin01: Yeah keywords are meant to improve ranking, but if we're having to alter them individually or cut out meaningful keywords to get better ranking, it seems like an issue with our retrieval algorithm that we would want to be aware of. I'll set up a meeting with Joe around some of these issues.

embedding-generation/eval_questions.json
NeethuESim: Jira is created to track this - https://jira.arm.com/browse/STESOL-575
I am collecting other examples, will update jira. Will let this question fail.

brikin01 (approved):
LGTM! Thanks for all the work and diagnosis!

Any follow-up items have been tracked in Jira tickets. Keywords can always be expanded/modified at a later date as needed for improved retrieval, but seems like we're getting good enough performance for most eval questions.

PR #114, opened 2026-07-16
Update Github Actions versions and add eval acceptable source

brikin01:
This PR does two things:
- Updates Github Actions workflow versions to higher versions which use Node 24, since [Node 20 is now deprecated](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners). This was showing warnings on all of our workflows
- Adds an acceptable source to a currently failing eval question about nginx performance tweaks; also adding same source to integration test revolving around the same query
- Corrects an ampere source URL to new lowercase version

### DIFF
--- a/.github/workflows/build-embeddings.yml
+++ b/.github/workflows/build-embeddings.yml
@@ -13,22 +13,22 @@ jobs:
       packages: write
     steps:
       - name: Checkout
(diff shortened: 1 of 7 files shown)


NeethuESim (approved):
LGTM

PR #116, opened 2026-07-20
Add eval questions for developer.arm.com content

jaidev17:
Add one question per content file added from developer.arm.com. All questions are giving hits for the expected URLs.

### DIFF
--- a/embedding-generation/eval_questions.json
+++ b/embedding-generation/eval_questions.json
@@ -2586,9 +2586,489 @@
     ]
   },
   {
-    "question": "devices that support SME and SME2",
+    "question": "Which phones, Macs, and iPads have native SME2 support?",
     "expected_urls": [
       "https://learn.arm.com/learning-paths/cross-platform/multiplying-matrices-with-sme2/1-get-started/#devices"
+    ]
+  },
+  {
+    "question": "Why would I use SME for matrix-heavy code?",
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109974/0100/Overview-of-SME?lang=en"
+    ]
+  },
+  {
+    "question": "Can you explain the difference between SME and SME2?",
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109246/0101/SME-Overview/SME-and-SME2?lang=en"
+    ]
+  },
+  {
+    "question": "What do I need to know about saving and restoring SME state?",
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109246/0101/SME-Overview/SME-context-save-restore?lang=en"
+    ]
+  },
+  {
+    "question": "How does ZA storage work in SME?",
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109246/0101/SME-Overview/SME-ZA-storage?lang=en"
+    ]
+  },
+  {
+    "question": "How do I check if SME or SME2 is supported?",
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109246/0101/SME-Overview/SME-and-SME2/If-SME-and-SME2-are-supported?lang=en"
+    ]
+  },
+  {
+    "question": "How do I build and run a simple SME application?",
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109246/0101/Toolchains-and-model-support/How-to-run-an-SME-application?lang=en"
+    ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


embedding-generation/eval_questions.json
```diff
@@ -2585,14 +2582,451 @@
...
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109246/0101/matmul-fp32--Single-precision-matrix-by-matrix-multiplication/preprocess-l-code?lang=en"
+    ]
+  },
+  {
+    "question": "What is the preprocess_r code for int8-to-int32 SME matrix-by-matrix multiplication?",
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109246/0101/matmul-int8--8-bit-integer-to-32-bit-integer-matrix-by-matrix-multiplication/preprocess-r-code?lang=en",
+      "https://developer.arm.com/documentation/109246/0101/matmul-int8--8-bit-integer-to-32-bit-integer-matrix-by-matrix-multiplication?lang=en"
+    ]
```
brikin01: Given only the first link actually contains the code, should we remove the second link since it's not a very strong hit?

embedding-generation/eval_questions.json
```diff
@@ -2585,14 +2582,451 @@
...
+      "https://developer.arm.com/documentation/109246/0101/Introduction/The-Scalable-Matrix-Extensions/Streaming-SVE-mode-and-ZA-storage?lang=en"
+    ]
+  },
+  {
+    "question": "What is the preprocess_l code for int8-to-int32 SME matrix-by-matrix multiplication?",
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109246/0101/matmul-int8--8-bit-integer-to-32-bit-integer-matrix-by-matrix-multiplication/preprocess-l-code?lang=en",
+      "https://developer.arm.com/documentation/109246/0101/matmul-int8--8-bit-integer-to-32-bit-integer-matrix-by-matrix-multiplication?lang=en"
+    ]
+  },
```
brikin01, line 2877: Similar question to above

embedding-generation/eval_questions.json
```diff
@@ -2585,14 +2582,451 @@
...
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109974/0100/Basic-SME-example/Assembly-code?lang=en"
+    ]
+  },
+  {
+    "question": "What is the preprocess_l function?",
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109246/0101/matmul-int8--8-bit-integer-to-32-bit-integer-matrix-by-matrix-multiplication/preprocess-l-function-overview?lang=en"
+    ]
+  },
```
brikin01: preprocess_l exists in both the FP32 and int8 examples. Is this intended to test a generic user query? If so, both exact preprocess_l pages seem like valid expected sources (ideally both should be returned in the top5, but our eval doesn't currently test that). If it’s intended to cover the int8 page specifically, could we add “in the int8-to-int32 matrix multiplication example” to remove the ambiguity?

embedding-generation/eval_questions.json
```diff
@@ -2585,14 +2582,451 @@
...
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109246/0101/Toolchains-and-model-support/Compiler-support?lang=en"
+    ]
+  },
+  {
+    "question": "What is the preprocess_r function?",
+    "expected_urls": [
+      "https://developer.arm.com/documentation/109246/0101/matmul-int8--8-bit-integer-to-32-bit-integer-matrix-by-matrix-multiplication/preprocess-r-function-overview?lang=en"
+    ]
+  },
```
brikin01: Similar question as to the `preprocess_l` query

brikin01 (commented):
#104 added 80 source URLs, while this PR adds 70 questions. All 80 URLs are referenced, but 22 appear only as alternatives in multi-URL questions. Because retrieving any expected URL counts as a pass, those 22 URLs are not independently exercised. This provides topical coverage rather than one retrieval check per source URL.

I don’t think that’s necessarily a problem or should block this PR, especially while we align on smoke-test versus benchmark goals. I just wanted to record the distinction for future discussion.

I also added a few minor comments/questions about the preprocess_l and preprocess_r queries. Looks good overall.

brikin01 (approved):

PR #124, opened 2026-07-28
Add logging for arm mcp server testing

jasonrandrews:


mcp-local/utils/invocation_logger.py
```diff
@@ -23,38 +24,77 @@
 
 
 LOG_FILE_NAME = "invocation_reasons.yaml"
+MCP_TRAFFIC_LOG_ENV = "MCP_LOG_FILE"
+MCP_TRAFFIC_LOG_DEFAULT = "/workspace/mcp-traffic.jsonl"
```
brikin01, line 28: Should rich JSONL logging be opt-in through MCP_LOG_FILE rather than enabled by default? The current default creates `mcp-traffic.jsonl` in the host project directory mounted at `/workspace`. Default-on is more convenient, so I could see either approach, but it seems worth considering and documenting explicitly.

mcp-local/server.py
```diff
@@ -67,7 +67,9 @@ def knowledge_base_search(query: str, invocation_reason: Optional[str] = None) -
         List of dictionaries with metadata including url and text snippets.
     """
     try:
-        return arm_kb_search.search(query, SEARCH_RESOURCES)
+        results = arm_kb_search.search(query, SEARCH_RESOURCES)
+        log_tool_result(entry_id, "knowledge_base_search", results)
```
brikin01, line 71: This currently writes JSONL call entries for all tools, but writes a corresponding result entry only for successful `knowledge_base_search` calls. Is that intentional? It seems we should either log results for all tools or scope JSONL logging entirely to `knowledge_base_search`.

PR #129, opened 2026-08-05
feat(ste-576): add gh action for OSSF Scorecard

NeethuESim:
GitHub Action for OpenSSF Scorecard - https://github.com/ossf/scorecard-action
Assess open source projects for security risks through a series of automated checklist - https://openssf.org/projects/scorecard/


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,81 @@
...
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+      # Uncomment the permissions below if installing in a private repository.
+      # contents: read
+      # actions: read
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
brikin01: This version uses deprecated Node 20, can we bump to a later one like v7?

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,81 @@
...
+        with:
+          persist-credentials: false
+
+      - name: "Run analysis"
+        uses: ossf/scorecard-action@f49aabe0b5af0936a0987cfb85d86b75731b0186 # v2.4.1
```
brikin01: Can we upgrade this action to the latest version, v2.4.4, retaining a SHA pin? It includes fixes that reduce false positives in the Pinned-Dependencies check, particularly for Dockerfiles using `FROM` scratch or pinned images supplied through build arguments.

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,81 @@
...
+          #     of the value entered here.
+          publish_results: true
+
+          # (Optional) Uncomment file_mode if you have a .gitattributes with files marked export-ignore
+          # file_mode: git
+
+      # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
+      # format to the repository Actions tab.
+      - name: "Upload artifact"
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4.6.1
```
brikin01: Similar to above, this action version uses deprecated Node20. Bump up to v7?

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,81 @@
...
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecard analysis
+    runs-on: ubuntu-latest
+    # `publish_results: true` only works when run from the default branch. conditional can be removed if disabled.
+    if: github.event.repository.default_branch == github.ref_name || github.ref_name == 'secure-build' || github.ref_name == 'stesol-576-ossf-scorecard' || github.event_name == 'pull_request'
```
brikin01: Do we need this condition? It looks like all of the configured triggers already satisfy it, so I think the job would behave the same without it. The explicit feature-branch check also seems like something we wouldn’t want to keep after the PR is merged.

brikin01 (changes requested):
Thanks for adding this scorecard, I think it will be useful!

I added a few comments around bumping versions to their latest.

Also, do we need all of these template comments? A lot of them repeat the upstream docs or describe options we aren’t using, and they make the workflow a bit harder to scan. Could we keep just the comments that explain repo-specific or security-related choices?

.github/workflows/scorecard.yml
NeethuESim: Yes, we can remove this since push works only on main.

NeethuESim:
> Thanks for adding this scorecard, I think it will be useful!
> 
> I added a few comments around bumping versions to their latest.
> 
> Also, do we need all of these template comments? A lot of them repeat the upstream docs or describe options we aren’t using, and they make the workflow a bit harder to scan. Could we keep just the comments that explain repo-specific or security-related choices?

I will remove any redundant comments. However, would like to keep the other options not used here along with their comments, as it would help later if we decide to change something. We know these options exist & could go search for them in their documentation directly. I understand its not neat, but atleast we know an option exists because of these commented lines of code.

brikin01 (approved):
LGTM!

PR #131, opened 2026-08-05
feat(stesol-580): add blackduck workflow

NeethuESim:
Add blackduck workflow. Scan resulting from PR merge will be available at Arm BlackDuck UI - https://arm.app.blackduck.com/api/risk-profile-dashboard

### DIFF
--- a/.github/workflows/black-duck-security-scan-ci.yml
+++ b/.github/workflows/black-duck-security-scan-ci.yml
@@ -24,7 +24,7 @@ jobs:
     permissions:
       contents: read
       pull-requests: write
-      security-events: write
+      security-events: read
       actions: read
 
     steps:
--- a/.github/workflows/black-duck-security-scan-ci.yml
+++ b/.github/workflows/black-duck-security-scan-ci.yml
@@ -42,7 +42,7 @@ jobs:
 
           ### SCANNING: Optional fields
           # blackducksca_scan_failure_severities: 'BLOCKER,CRITICAL'
-          detect_args: "--detect.accuracy.required=LOW"
+          detect_args: "--detect.accuracy.required=NONE"
 
           ### FIX PULL REQUEST CREATION: Uncomment below to enable
           blackducksca_fixpr_enabled: true
@@ -75,7 +75,7 @@ jobs:
           blackducksca_url: ${{ vars.BLACKDUCKSCA_URL }}
           blackducksca_token: ${{ secrets.BLACKDUCKSCA_TOKEN }}
           blackducksca_scan_full: false
-          detect_args: "--detect.accuracy.required=LOW"
+          detect_args: "--detect.accuracy.required=NONE"
 
           ### Mark build status if policy violating issues are found
           mark_build_status: 'success'
--- a/.github/workflows/black-duck-security-scan-ci.yml
+++ b/.github/workflows/black-duck-security-scan-ci.yml
@@ -29,7 +29,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     permissions:
-      contents: write
+      contents: read
       pull-requests: write
       security-events: write
       actions: read
@@ -68,7 +68,8 @@ jobs:
 
           ### FIX PULL REQUEST CREATION
           blackducksca_fixpr_enabled: true
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          # Dedicated token with Contents and Pull requests read/write access.
+          github_token: ${{ secrets.BLACKDUCK_GITHUB_TOKEN }}
 
           ### SARIF report generation for upload to GitHub Advanced Security
           blackducksca_reports_sarif_create: true
--- a/.github/workflows/black-duck-security-scan-ci.yml
+++ b/.github/workflows/black-duck-security-scan-ci.yml
@@ -55,10 +55,11 @@ jobs:
           blackducksca_prcomment_enabled: true 
           # github_token: ${{ secrets.GITHUB_TOKEN }} # Required when PR comments is enabled
 
-          ### SARIF report generation and upload to GitHub Adavanced Security: Uncomment below to enable
-          blackducksca_reports_sarif_create: true # Create Black Duck SCA SARIF report and upload it as artifact
-          blackducksca_upload_sarif_report: true  # Upload Black Duck SCA SARIF report in GitHub Advanced Security tab
-          # github_token: ${{ secrets.GITHUB_TOKEN }} # Required when blackducksca_upload_sarif_report is set as true
+          ### SARIF report generation for upload to GitHub Advanced Security
+          blackducksca_reports_sarif_create: true
+          blackducksca_reports_sarif_file_path: blackduck-results.sarif
+          ### Set upload to false & then later upload via github/codeql-action/upload-sarif@v4.37.4 to avoid token-permissions issue flagged by scorecard
+          blackducksca_upload_sarif_report: false
 
           ### Mark build status if policy violating issues are found
           mark_build_status: 'success'
@@ -69,6 +70,12 @@ jobs:
 
           ### Uncomment below configuration if Bridge diagnostic files needs to be uploaded
           include_diagnostics: true
+
+      - name: Upload Black Duck SARIF
+        if: ${{ github.event_name != 'pull_request' }}
+        uses: github/codeql-action/upload-sarif@f205ea1c3313d32999d8d6a48b4f6530d4437b38 # v4.37.4
+        with:
+          sarif_file: blackduck-results.sarif
           
       - name: Black Duck SCA PR Scan
         id: black-duck-pr-scan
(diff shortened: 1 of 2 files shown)


.github/workflows/black-duck-security-scan-ci.yml
```diff
@@ -0,0 +1,91 @@
...
+          # blackducksca_scan_failure_severities: 'BLOCKER,CRITICAL'
+          detect_args: "--detect.accuracy.required=NONE"
+
+          ### FIX PULL REQUEST CREATION: Uncomment below to enable
+          blackducksca_fixpr_enabled: true
+          github_token: ${{ secrets.GITHUB_TOKEN }} # Required when Fix PRs is enabled
+
+          ### PULL REQUEST COMMENTS: Uncomment below to enable
+          blackducksca_prcomment_enabled: true 
+          github_token: ${{ secrets.GITHUB_TOKEN }} # Required when PR comments is enabled
```
brikin01: I think this line needs to be deleted, github_token is already defined in line 52, so this creates a duplicate key and makes the workflow invalid (https://github.com/arm/mcp/actions/runs/32064906758)

.github/workflows/black-duck-security-scan-ci.yml
```diff
@@ -0,0 +1,91 @@
...
+    - cron: '34 5 * * 6'
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
```
brikin01, line 32: We have `blackducksca_fixpr_enabled: true` below, but for it to create PRs, I think we need this to have `contents: write` permission (looking at the sample workflow in https://docs.blackduck.com/r/bridge/latest/bridge-cli-guide/using-the-github-app-with-black-duck-sca.html). That or we should set the fixpr config to false.

.github/workflows/black-duck-security-scan-ci.yml
```diff
@@ -0,0 +1,91 @@
...
+          ### SCANNING: Optional fields
+          # blackducksca_scan_failure_severities: 'BLOCKER,CRITICAL'
+          detect_args: "--detect.accuracy.required=NONE"
+
+          ### FIX PULL REQUEST CREATION: Uncomment below to enable
+          blackducksca_fixpr_enabled: true
+          github_token: ${{ secrets.GITHUB_TOKEN }} # Required when Fix PRs is enabled
+
+          ### PULL REQUEST COMMENTS: Uncomment below to enable
+          blackducksca_prcomment_enabled: true
```
brikin01: To my understanding, this option enables comments on PRs, but this step explicitly doesn't run on PRs. Maybe this config should be moved to the `black-duck-pr-scan` step (along with the `github_token` line)?

.github/workflows/black-duck-security-scan-ci.yml
```diff
@@ -0,0 +1,91 @@
...
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main", "secure-build" ]
+  schedule:
+    - cron: '34 5 * * 6'
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
```
brikin01, line 28: I think we might need `DETECT_PROJECT_NAME` and `DETECT_PROJECT_VERSION_NAME` defined in the env here? Referring to quickstart: https://docs.blackduck.com/r/bridge/latest/bridge-cli-guide/quickstart-black-duck-security-scan-action-with-black-duck-sca.html. Or are these configured elsewhere?

brikin01 (changes requested):
Looks pretty good, just a few small changes/questions!

.github/workflows/black-duck-security-scan-ci.yml
NeethuESim: Good catch, I missed this.

.github/workflows/black-duck-security-scan-ci.yml
NeethuESim: I was thinking, won't opening a PR for fixing issue with fixpr, would need comment enabled right ? But we should also have it in the PR flow below.

.github/workflows/black-duck-security-scan-ci.yml
NeethuESim: You are right, this needs to be here.

.github/workflows/black-duck-security-scan-ci.yml
brikin01: To my understanding, I think the flow would be:
1. A non-PR Black Duck flow runs, for example on main. It doesn’t need PR comments enabled because there is no PR to comment on.
2. Black Duck finds an eligible issue and creates a Fix PR.
3. Creating that PR triggers the Black Duck PR workflow. Because the Fix PR was created using `GITHUB_TOKEN`, GitHub will likely leave the workflow run waiting for approval from someone with write access. Once approved, the PR scan runs with comments enabled.

So these are separate processes, and there isn’t anywhere for a PR comment to be added during the original non-PR scan.

NeethuESim: this makes sense, it will be separate flows.

.github/workflows/black-duck-security-scan-ci.yml
```diff
@@ -0,0 +1,106 @@
...
+          github_token: ${{ secrets.BLACKDUCK_GITHUB_TOKEN }}
+
+          ### SARIF report generation for upload to GitHub Advanced Security
+          blackducksca_reports_sarif_create: true
+          blackducksca_reports_sarif_file_path: blackduck-results.sarif
+          ### Set upload to false & then later upload via github/codeql-action/upload-sarif@v4.37.4 to avoid token-permissions issue flagged by scorecard
+          blackducksca_upload_sarif_report: false
+
+          ### Mark build status if policy violating issues are found
+          mark_build_status: 'success'
```
brikin01, line 76: Why do we mark the build status as success? I would think a violation would cause a failure?

.github/workflows/black-duck-security-scan-ci.yml
```diff
@@ -0,0 +1,106 @@
...
+# read our documentation at https://github.com/blackduck-inc/black-duck-security-scan
+
+name: CI Black Duck security scan
+
+on:
+  push:
+    branches: [ "main", "secure-build" ] # TODO: remove "secure-build" after merge into main
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main", "secure-build" ]  # TODO: remove "secure-build" after merge into main
```
brikin01, line 17: We should be good to delete secure-build now (and also need to change the target branch of the PR to main)

.github/workflows/black-duck-security-scan-ci.yml
```diff
@@ -0,0 +1,106 @@
...
+          blackducksca_token: ${{ secrets.BLACKDUCKSCA_TOKEN }}
+          blackducksca_scan_full: false
+          detect_args: "--detect.accuracy.required=NONE"
+
+          ### PULL REQUEST COMMENTS
+          blackducksca_prcomment_enabled: true
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+          ### Mark build status if policy violating issues are found
+          mark_build_status: 'success'
```
brikin01, line 106: Same question as previous

.github/workflows/black-duck-security-scan-ci.yml
```diff
@@ -0,0 +1,106 @@
...
+
+          ### To enable Black Duck SCA policy badges
+          blackducksca_policy_badges_create: true
+          blackducksca_policy_badges_maxCount: 5
+
+          ### To upload Bridge diagnostic files
+          include_diagnostics: true
+
+      - name: Upload Black Duck SARIF
+        if: ${{ github.event_name != 'pull_request' }}
```
brikin01, line 86: If we do change policy violations to fail the scan, should this step use an always()/!cancelled() condition so the SARIF report is still uploaded and the findings remain visible?

brikin01 (changes requested):
Thanks for all the changes, seems to be in a pretty good spot!

I left a few comments about policy-failure behavior and cleaning up the temporary secure-build configuration.

This looks like it covers the initial Black Duck setup shared by STESOL-576 and STESOL-580. I’m assuming the release-evidence side of STESOL-580, e.g. generating the complete SBOM for what we actually ship, will come in a follow-up PR tied into the release workflow. Is that what you had in mind?

NeethuESim:
> Thanks for all the changes, seems to be in a pretty good spot!
> 
> I left a few comments about policy-failure behavior and cleaning up the temporary secure-build configuration.
> 
> This looks like it covers the initial Black Duck setup shared by STESOL-576 and STESOL-580. I’m assuming the release-evidence side of STESOL-580, e.g. generating the complete SBOM for what we actually ship, will come in a follow-up PR tied into the release workflow. Is that what you had in mind?

Yes, this PR only covers the initial integration. Will open incremental PRs for the remaining items.

.github/workflows/black-duck-security-scan-ci.yml
NeethuESim: You are right, I will add the failure condition based on severity detected.

.github/workflows/black-duck-security-scan-ci.yml
NeethuESim: I wasn't sure on this, let me check in slack. I can then directly create this PR against main instead of merge into secure build

.github/workflows/black-duck-security-scan-ci.yml
NeethuESim: Yes, will make sure we have report generated even after policy violation. I will open another PR for the policy violations.

NeethuESim:
Closing this PR to open it against main

PR #132, opened 2026-08-06
Add dependabot version scanning config

brikin01:
Adds Dependabot version scanning for Github Actions workflows and our various `uv`/`pip` environments. This is in addition to the security scanning that is already present. To reduce the amount of PR noise, I've set it to run monthly and group all updates into one PR for each package manager environment. We can easily change this to more or less often as desired in the future.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
     # A root entry scans workflows under .github/workflows.
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
     groups:
       github-actions:
         patterns:
@@ -14,7 +14,7 @@ updates:
   - package-ecosystem: "uv"
     directory: "/embedding-generation"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
     groups:
       embedding-dependencies:
         patterns:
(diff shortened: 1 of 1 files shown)


NeethuESim (commented):
@brikin01 one danger I see with this version package update is regression issues. Existing flows might break because of incompatibility between packages after upgrade. Can you add some dependency here that update should be made only after the integration & unit tests pass with these changes ? Or if these changes are made after PR creation, then merge should be allowed only if all the integration & unit tests pass.

brikin01:
> @brikin01 one danger I see with this version package update is regression issues. Existing flows might break because of incompatibility between packages after upgrade. Can you add some dependency here that update should be made only after the integration & unit tests pass with these changes ? Or if these changes are made after PR creation, then merge should be allowed only if all the integration & unit tests pass.

The typical flow is:
1. Dependabot runs on a weekly cadence
2. If it detects a version can/should be updated, it will open a PR for those updates, thus triggering any CI workflows we have.
3. We still then need to manually review and merge it.

So I think there shouldn't be a problem there, in that our tests should catch any regressions before we ever merge. And from there we could either close the Dependabot PR or figure out how to use the latest version without breaking CI. Any rules around requiring tests to pass before merging will just be a matter of having stricter rules on our main branch, which I believe is part of the work of [STESOL-557](https://jira.arm.com/browse/STESOL-557).

NeethuESim:
> [STESOL-557](https://jira.arm.com/browse/STESOL-557)

cool. [STESOL-557](https://jira.arm.com/browse/STESOL-557) will prevent any accidental merging for any PRs without the integration/unit tests passing.

JoeStech (approved):

PR #135, opened 2026-08-11
STESOL-570 pin build inputs

JoeStech:
This branch implements STESOL-570 by making the MCP image build reproducible and offline during installation.

Top-level changes:

- Pins Ubuntu, embeddings, Performix, migrate-ease, Python packages, and OS packages by version and digest/hash.
- Replaces the broad arm-linux-migration-tools installer with only the required tools.
- Creates verified AMD64 and Arm64 wheel and .deb bundles.
- Publishes those inputs as a private, multi-architecture GHCR artifact.
- Makes the final Dockerfile consume that artifact by immutable digest.
- Installs apt and Python dependencies without network access.
- Packages the embedding model, metadata, and index locally.
- Adds workflows, lock manifests, tests, and documentation for updating, publishing, inspecting, and consuming build inputs.


.github/workflows/build-mcp-image.yml
```diff
@@ -111,6 +112,13 @@ jobs:
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
 
+      - name: Log in to GHCR for locked build inputs
+        uses: docker/login-action@v4
```
NeethuESim: this will get flagged by security as not pinned by hash. latest is v4.6.0, so can be changed to docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0 https://github.com/docker/login-action/commit/dbcb813823bdd20940b903addbd779551569679f

.github/workflows/integration-tests.yml
```diff
@@ -46,20 +50,24 @@ jobs:
...
-        run: |
-          python -m pip install --upgrade pip
-          pip install -r mcp-local/tests/requirements.txt
+        run: python -m pip install -r mcp-local/tests/requirements.txt
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
 
+      - name: Log in to GHCR for locked build inputs
+        uses: docker/login-action@v4
```
NeethuESim: same comment to pin with hash

mcp-local/requirements.txt
brikin01, line 1: Do we need this anymore?

.github/workflows/integration-tests.yml
```diff
@@ -46,20 +50,24 @@ jobs:
           cache: pip
 
       - name: Install test dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install -r mcp-local/tests/requirements.txt
+        run: python -m pip install -r mcp-local/tests/requirements.txt
```
brikin01: Instead of using a requirement.txt inside of tests, should we just use a dev group inside of `mcp-local/pyproject.toml`?

mcp-local/requirements.lock
brikin01, line 1: Do we want to keep this duplicate dependency list? We could either keep the pip-compatible `requirements.lock`, generate it from `uv.lock` during input acquisition (and not have it checked in), or adopt the embedding build’s `uv sync --locked` approach. Do you have a preferred approach?

.github/workflows/integration-tests.yml
```diff
@@ -23,7 +24,10 @@ jobs:
         run: python -m pip install pytest
 
       - name: Run unit tests
-        run: python -m pytest -q mcp-local/tests/test_invocation_logger.py
+        run: |
+          python -m pytest -q \
+            mcp-local/tests/test_build_inputs.py \
+            mcp-local/tests/test_invocation_logger.py
```
brikin01: Why is test_apx_mounts.py omitted from the unit-test job? Should we include it explicitly, or separate unit and integration tests so the complete unit suite can run automatically?

mcp-local/Dockerfile
```diff
@@ -14,124 +14,112 @@
 
 # syntax=docker/dockerfile:1.7
```
brikin01, line 15: I didn't do this in embedding-generation either, but theoretically this could also be pinned to a hash. https://docs.docker.com/build/buildkit/frontend/

mcp-local/.python-version
```diff
@@ -0,0 +1 @@
+3.12
```
brikin01, line 1: Here we use 3.12, in embedding-generation I used 3.13. We don't necessarily need them to be the same, but should we standardize?

.github/workflows/build-mcp-inputs.yml
```diff
@@ -0,0 +1,159 @@
...
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Set up Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.12"
```
brikin01, line 37: Should this use an exact Python patch version or a digest-pinned Python image rather than resolving the latest 3.12 on each run?

.github/workflows/build-mcp-inputs.yml
```diff
@@ -0,0 +1,159 @@
...
+          username: ${{ github.actor }}
+          password: ${{ github.token }}
+
+      - name: Verify existing package is private
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: bash .github/scripts/verify-ghcr-package-private.sh "${PACKAGE_NAME}" --allow-missing
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4
```
brikin01, line 77: [Non-blocking/possibly optional] This pins the setup action but apparently [we could also pin Buildx/BuildKit versions ](https://docs.docker.com/build/ci/github-actions/configure-builder/#version-pinning)(I didn't do this in the embedding workflows). Thoughts on whether we want to worry about that?

.github/workflows/build-mcp-image.yml
```diff
@@ -2,7 +2,7 @@ name: Build MCP Image
 
 on:
   workflow_run:
-    workflows: ["Build Embeddings Image"]
+    workflows: ["Build Offline Embedding Pipeline"]
```
brikin01: So this triggers right after offline embedding pipeline completes, but we won't end up using the latest embeddings, since we pin to a specific image inside the dockerfile. Given that, how do we want to handle the relationship between these two and the full release process/frequency going forward?

mcp-local/Dockerfile
```diff
@@ -14,124 +14,112 @@
...
 
-ARG EMBEDDINGS_IMAGE=armlimited/arm-mcp:embeddings-latest
-ARG EMBEDDING_MODEL=all-MiniLM-L6-v2
-# EMBEDDINGS_IMAGE must point to an embeddings image tag (e.g., armlimited/arm-mcp:embeddings-YYYY-MM-DD).
+# Keep these defaults synchronized with mcp-local/build-inputs.lock.json.
+ARG EMBEDDINGS_IMAGE=ghcr.io/arm/mcp-embedding-vectorstore@sha256:724a53a6a8b29893c196c406a6ba9a41a1ec3b122b7aa2ceb587a67991519f88
+ARG MCP_BUILD_INPUTS_IMAGE=ghcr.io/arm/mcp-build-inputs@sha256:d38c64ad12493ddacfe7a99e49f47e6cfcf1f9d2acf1696c797645d3099758a3
+ARG UBUNTU_IMAGE=ubuntu:24.04@sha256:4fbb8e6a8395de5a7550b33509421a2bafbc0aab6c06ba2cef9ebffbc7092d90
```
brikin01, line 20: Could the release workflow load and validate these image references from build-inputs.lock.json, rather than duplicating them here? That would make the lock the single source of truth.

mcp-local/build-inputs.lock.json
```diff
@@ -0,0 +1,336 @@
...
+      "arm64"
+    ],
+    "verification": "source archive SHA256"
+  },
+  "performix": {
+    "version": "1.0.0",
+    "artifacts": {
+      "amd64": {
+        "url": "https://artifacts.tools.arm.com/arm-performix/cli/1.0.0/ArmPerformix-cli-1.0.0-linux-x64.tar.gz",
+        "sha256": "83edba4708daf6633b4d5a85270d14ee2913c9bfc1de523ab9280b0d9ac5461e"
```
brikin01, line 324: Is this SHA256 provided by the Performix team, or calculated from the downloaded artifact?

mcp-local/build-inputs.lock.json
```diff
@@ -0,0 +1,336 @@
...
+            "tzdata_2026c-0ubuntu0.24.04.1_all.deb": "ef12c9ef81905b5ac558504b2f8c20da9be5f28861020b3fb5e8b15fb3fae2f6"
+          }
+        }
+      }
+    },
+    "source": "Ubuntu snapshot service",
+    "verification": "signed snapshot metadata plus checked-in SHA256 for every downloaded deb"
+  },
+  "migrate_ease": {
+    "repository": "https://github.com/JoeStech/migrate-ease",
```
brikin01: Not necessarily relevant to this PR, but just curious why we have this fork of migrate-ease vs the original repo?

brikin01 (changes requested):
We're coming along nicely, this was a big chunk and great improvement to the build flow! My main concern is that the MCP build still triggers immediately after embedding generation while consuming the previously pinned embedding digest, so we should clarify the promotion and release flow. I also left a few smaller comments and questions.

mcp-local/.python-version
JoeStech: let's leave as-is for now and I'll put a backlog item to standardize on the next LTS Ubuntu release. 24.04 doesn't have 3.13, and when I looked into it it would have been pretty messy to use 3.13 in Ubuntu 24.04 in a hermetic build process for very little benefit.

mcp-local/requirements.txt
JoeStech: I went back and forth on this because of the ease-of-use build question I put in Slack, but I think you're right in that we should remove it, and be consistent with the embeddings flow. One less point where things can diverge

.github/workflows/build-mcp-inputs.yml
JoeStech: good question, this particular one should be fine since it's not copied into the MCP image, it just runs the acquisition tooling. I'll keep it as-is just to remove a place where we'd have to do manual security patch upgrades

.github/workflows/build-mcp-inputs.yml
JoeStech: rare that this would have an impact, but it's a good callout and I've added it to a backlog ticket

mcp-local/Dockerfile
JoeStech: I put this in a backlog ticket

mcp-local/requirements.lock
JoeStech: I changed to remove this and have the source of truth be pyproject.toml and uv.lock, but the requirements.lock is included in the immutable GHCR artifact. Leaving uv sync out of the Dockerfile though

.github/workflows/build-mcp-image.yml
JoeStech: Great comment, I changed things so that the embedding build generates a PR instead of triggering the build. When the PR is merged, it will trigger the build with the new fixed embeddings.

.github/workflows/integration-tests.yml
JoeStech: removed the file entirely, it was outside scope.

mcp-local/Dockerfile
JoeStech: I changed it to get the passed hashes from build-inputs, however, left these for the case that people just want a simple build locally

.github/workflows/build-embeddings.yml
```diff
@@ -186,3 +190,90 @@ jobs:
...
+          ref: main
+          fetch-depth: 0
+
+      - name: Update checked-in embedding pin
+        id: update
+        shell: bash
+        run: |
+          set -euo pipefail
+          git checkout -B "${PROMOTION_BRANCH}"
+          python3 .github/scripts/update-mcp-embedding-pin.py \
```
brikin01, line 249: I think this file needs added to the branch

brikin01 (commented):
Really like the human-in-the-loop promotion flow! I think after missing script is added and the merge conflicts with #136 are resolved, it should be g2g.

mcp-local/build-inputs.lock.json
JoeStech: we calculate it. Thanks for mentioning because we say it's a vendor artifact when it's not. I changed that line, will push

mcp-local/build-inputs.lock.json
JoeStech: Missed this one. Good comment, need to change this. It was still pointing to my version because I had a fix, but the fix was upstreamed into migrate-ease repo long ago

brikin01 (approved):
LGTM! I saw your comment on migrate-ease, not sure if you wanted to add to this PR or just add as a follow-up, either makes sense to me

PR #146, opened 2026-08-20
Codeowners and release flow updates

JoeStech:
- Added CODEOWNERS: release/supply-chain files require @arm/mcp-security-team; general files remain unrestricted.
- Reworked releases around reviewed version PRs; removed self-merging, bump branches, date tags, and rerun/race logic.
- Embedding promotion PRs now include the next minor version.
- Manual workflow supports reviewed major, minor, and hotfix PRs.
- Merged version PRs build the exact commit, publish multi-architecture images, and create matching tags/releases.
- Added reusable version-bump tooling, documentation, and tests.

### DIFF
--- a/.github/workflows/build-mcp-image.yml
+++ b/.github/workflows/build-mcp-image.yml
@@ -2,7 +2,6 @@ name: Build MCP Image
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened]
     branches: [main]
     paths:
       - mcp-local/server.json
@@ -298,11 +297,10 @@ jobs:
           SOURCE_SHA: ${{ needs.validate-release.outputs.source_sha }}
         run: |
           set -euo pipefail
-          notes_file="${RUNNER_TEMP}/release-notes.md"
-          printf 'Docker image: `%s:%s`\n\nSource commit: `%s`\n' \
-            "${IMAGE}" "${VERSION}" "${SOURCE_SHA}" > "${notes_file}"
+          notes="$(printf 'Docker image: `%s:%s`\n\nSource commit: `%s`' \
+            "${IMAGE}" "${VERSION}" "${SOURCE_SHA}")"
           gh release create "v${VERSION}" \
             --target "${SOURCE_SHA}" \
             --title "v${VERSION}" \
             --generate-notes \
-            --notes-file "${notes_file}"
+            --notes "${notes}"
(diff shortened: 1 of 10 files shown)


.github/workflows/embedding-unit-tests.yml
brikin01, line 1: I haven't looked at these changes, but I did see you had modified `embedding-unit-tests.yml`. You had added this workflow as a required check for the repo, but it doesn't always run on every PR (only when certain paths are modified; see #132 for an example where it's required but not triggered). I was in the process of removing the paths and having it run unconditionally on every PR. Thoughts about just adding that change here? I also was thinking of renaming the `job` field to a more descriptive name than `test` (although that's much less important, especially since the main workflow name is still informative). Happy to also change this in a different PR, it was just top of mind.

JoeStech: Yes, I'll make these changes here. I'll call it embedding-unit-tests instead of test as well.

JoeStech: OK this is done, very small change

.github/workflows/build-mcp-image.yml
```diff
@@ -2,58 +2,162 @@ name: Build MCP Image
 
 on:
   pull_request:
-    types: [closed]
+    types: [opened, synchronize, reopened]
```
brikin01: (Non-blocking/nit) I think these are the default types (https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request) , so this could be deleted

.github/workflows/embedding-unit-tests.yml
```diff
@@ -1,22 +1,20 @@
...
-      - 'embedding-generation/**'
-      - '.github/workflows/embedding-unit-tests.yml'
+    branches: [main, secure-build]
 
 permissions:
   contents: read
 
 jobs:
-  test:
+  embedding-unit-tests:
```
brikin01, line 17: Just need to update the ruleset once this is merged I think

brikin01 (commented):
Looks good to me, just had one minor nit that you can take or leave. I think just resolving the Copilot comments and solving the merge conflict (which I think will let the tests complete) will make it g2g. Then happy to approve the final commit!

.github/workflows/embedding-unit-tests.yml
brikin01: I updated the ruleset FYI, so it should be good to go

brikin01 (approved):
LGTM

PR #151, opened 2026-08-21
Trigger input bundle updates

brikin01:
**Summary**

- Automatically rebuilds MCP input bundles and the embedding toolchain when their source inputs change.
- Opens reviewed PRs to update the resulting immutable image pins.
- Prevents stale builds from being promoted if main advances.
- Documents automatic and manual input-refresh workflows.

Changes within `mcp-local/build-inputs.lock.json`—such as updating the Ubuntu package snapshot, Performix artifact, or migrate-ease revision—are intentionally not automatic because pin updates would cause a rebuild loop. These require a manual workflow run on main.

### DIFF
--- a/.github/scripts/update-embedding-toolchain-pin.py
+++ b/.github/scripts/update-embedding-toolchain-pin.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+"""Update the immutable embedding toolchain input used by the pipeline."""
+
+from __future__ import annotations
+
+import argparse
+import json
+from pathlib import Path
+import re
+
+
+REPOSITORY = Path(__file__).resolve().parents[2]
+DEFAULT_LOCK = REPOSITORY / "embedding-generation" / "pipeline-inputs.lock.json"
+REFERENCE_PATTERN = re.compile(
+    r"^ghcr\.io/[A-Za-z0-9_.-]+/mcp-embedding-generator@sha256:[0-9a-f]{64}$"
+)
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--reference", required=True)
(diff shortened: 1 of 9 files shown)


.github/scripts/update-embedding-toolchain-pin.py
```diff
@@ -0,0 +1,46 @@
...
+def update_pin(*, reference: str, lock_file: Path) -> None:
+    if not REFERENCE_PATTERN.fullmatch(reference):
+        raise ValueError(
+            f"invalid immutable embedding toolchain reference: {reference}"
+        )
+
+    lock = json.loads(lock_file.read_text(encoding="utf-8"))
+    if set(lock) != {"generator_image", "intrinsic_chunks_image"}:
+        raise ValueError("pipeline input lock does not contain the expected keys")
+
```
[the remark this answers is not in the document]
brikin01: Keeping the exact-key validation intentionally. `build-embeddings.yml` also requires the lockfile to contain only these two keys, so loosening only the updater would make the validations inconsistent. If we add metadata later, we should update the lockfile schema, consumer validation, and tests together.

.github/workflows/build-mcp-inputs.yml
```diff
@@ -151,3 +194,60 @@ jobs:
             echo ""
             echo "Record this digest in \`mcp-local/build-inputs.lock.json\` before the release workflow consumes it."
```
JoeStech: I think we can remove this now, since the workflow automatically opens a promotion PR. Don't want folks to do extra work they don't need to do or get confused.

JoeStech (changes requested):

.github/workflows/build-mcp-inputs.yml
brikin01: Good call, removed in b6dd56ffd806e702cf7705ca64d83a37318009f7

JoeStech (approved):

PR #158, opened 2026-08-25
fix: disable Black Duck SCA policy badges creation

NeethuESim:
Disabling the policy badge creation for clean scan since its failing due to insufficient permissions for PAT token to modify repo ReadMe file to add the badge.


brikin01 (approved):
LGTM

PR #164, opened 2026-08-26
remove performix from MCP Server and documentation in this repo

JoeStech:
- Removed bundled Performix runtime, apx_recipe_run, APX utilities, queries, tests, and CI/SSH setup.
- Removed Performix from build inputs, Docker packaging, server registry metadata, and in-repo setup docs.
- Added the performance-routing hint to knowledge_base_search.
- Updated client prompts to use knowledge search, then the dedicated Performix MCP server, with Linux perf fallback.

NOTE: there are some upstream registry and documentation changes that still need to be done outside this repo, which I will do separately.


agent-integrations/vs-code/arm-enablement.prompt.md
```diff
@@ -6,10 +6,10 @@ name: 'arm-enablement'
 description: 'Assess an OSS codebase with Arm MCP and generate a professional Arm enablement report as Markdown and PDF'
 argument-hint: '[local workspace or GitHub repo URL] [--apply-fixes optional]'
 agent: 'agent'
-tools: ['search/codebase', 'search/fileSearch', 'search/textSearch', 'search/listDirectory', 'edit/editFiles', 'execute/runInTerminal', 'execute/getTerminalOutput', 'read/terminalLastCommand', 'arm-mcp/skopeo', 'arm-mcp/check_image', 'arm-mcp/knowledge_base_search', 'arm-mcp/migrate_ease_scan', 'arm-mcp/mca', 'arm-mcp/apx_recipe_run', 'arm-mcp/sysreport_instructions']
+tools: ['search/codebase', 'search/fileSearch', 'search/textSearch', 'search/listDirectory', 'edit/editFiles', 'execute/runInTerminal', 'execute/getTerminalOutput', 'read/terminalLastCommand', 'arm-mcp/skopeo', 'arm-mcp/check_image', 'arm-mcp/knowledge_base_search', 'arm-mcp/migrate_ease_scan', 'arm-mcp/mca', 'arm-mcp/sysreport_instructions']
```
brikin01: Should we revisit `tools:` for all of these VSCode prompts? To my understanding, in VS Code, the prompt’s [tool list takes precedence](https://code.visualstudio.com/docs/agent-customization/prompt-files#_tool-list-priority) over default agent tools, so installed Performix tools and/or the terminal needed for the `perf` fallback may be unavailable.

Possible options:

- Omit `tools:` entirely to inherit all enabled Agent/MCP tools.
- Add `performix/*` and the terminal tool explicitly (assuming that's the name of the MCP)
- Omit `tools:` and use [`#tool:` references](https://code.visualstudio.com/docs/agent-customization/prompt-files#_prompt-file-format) in the body to recommend tools without restricting others.

agent-integrations/codex/arm-hotspots-optimization.md
```diff
@@ -19,9 +20,9 @@ Steps to follow:
         * For Python code: `python /home/user/workspace/train.py`
         * For Java: `java -cp "absolute/path/to/class" some.package.Main`
     * If localhost is requested, use `localhost` as the remote_ip.
```
brikin01: Some of this guidance seems closely tied to the removed `apx_recipe_run` interface, including `cmd` and `remote_ip`. Would it be better for these prompts to describe the required outcome (i.e. identify the workload, select or configure an appropriate target, and run the recipe) while deferring the exact parameters and target workflow to the connected Performix tooling?

agent-integrations/vs-code/arm-enablement.prompt.md
JoeStech: good call, removed tools list

brikin01 (commented):
Looks pretty good, even more removed code/files now!

I think you missed a few files to delete under vs-code and gemini though. Once those are gone, happy to approve!

brikin01 (approved):
LGTM!

PR #169, opened 2026-09-01
feat(stesol-576-590): Enable BlackDuck Policy Gating + Scan build-inputs & embedding ghcr images

NeethuESim:
### PR Description

This PR covers changes for [STESOL-576](https://jira.arm.com/browse/STESOL-576) (Implement Policy gating)  & [STESOL-590](https://jira.arm.com/browse/STESOL-590) (Schedule Blackduck Scan for GHCR build-input and embedding images)

- Release policy is defined with scan failing when Blocker/Critical issues are found via `blackducksca_scan_failure_severities: 'BLOCKER,CRITICAL'` tags. Currently it is setup for license & security issues. Unable to test the license only blocking because of blackduck server issues.

- All issues are tracked via Arm BlackDuck UI dashboard - https://arm.app.blackduck.com/api/projects/53e38e5f-dafd-4b59-98cd-b73c2455d09e

- Policy violation is also set in Arm BlackDuck UI dashboard using tag `applyhighvulnspolicyasblocker` 

- Since the build-embeddings generates a scratch image with data-only artifacts rather than runnable containers, Blackduck finds no components/vulnerabilities. For more vulnerability coverage, embedding-toolchain is also scanned.

- Scan is setup for following versions under Arm:MCP project -
[mcp-source-1.0](https://arm.app.blackduck.com/api/projects/53e38e5f-dafd-4b59-98cd-b73c2455d09e/versions/4b4105cc-75bc-4cf2-a124-2a7f7dc4e3ee/components) - Codebase Source Scan
[mcp-build-inputs-container-amd64-1.0](https://arm.app.blackduck.com/api/projects/53e38e5f-dafd-4b59-98cd-b73c2455d09e/versions/985385d4-b218-4ec3-8964-2b01b7182f12/components) - GHCR build-inputs for amd64
[mcp-build-inputs-container-arm64-1.0](https://arm.app.blackduck.com/api/projects/53e38e5f-dafd-4b59-98cd-b73c2455d09e/versions/5e87011c-8a27-4dc2-994c-9ef0685416b3/components) - GHCR build-inputs for arm64
[mcp-embedding-vectorstore-1.0](https://arm.app.blackduck.com/api/projects/53e38e5f-dafd-4b59-98cd-b73c2455d09e/versions/f49a073e-5564-42e2-8281-c61e6087261b/components)  - GHCR embedding-vectorstore 
[mcp-embedding-generator-container-1.0](https://arm.app.blackduck.com/api/projects/53e38e5f-dafd-4b59-98cd-b73c2455d09e/versions/22e86295-f2f5-481d-b765-86f1259cc03a/components) - GHCR embedding-generator-container

- SBOM reports for all these can be generated & downloaded from the UI in CycloneDX format. SBOM for the runtime containers It is also uploaded as release artifact & added to the image.

- Dry run is added for testing purpose. We can keep it to debug BlackDuck issues in future or remove it. I am good with either.

- Automatic main or scheduled runs publish only after successful scans.

- Scans for the source code is triggered through `black-duck-security-scan-ci.yml` where as scan for build-inputs & embedding images is through `action.yml `invoked from `build-mcp-inputs.yml`, `build-embeddings.yml` & `build-embedding-toolchain.yml`. Two flows are kept separate since they scan different targets.

**- Currently BlackDuck Scans are failing due to timeout issue with their servers. I have reached out to their team via slack.**

### Scan Architecture & Flow

Repository
├-- Source scan
│  └─ Black Duck source/SCA analysis
│
└─ Container build workflows
   ├-- MCP build-input images: AMD64 + ARM64
   ├-- Embedding toolchain: ARM64
   └─ Embedding vectorstore: ARM64
          ↓
      Shared container-scan action
          ↓
      Policy gate
          ↓
      Publish + provenance

**Source-code scan -** [black-duck-security-scan-ci.yml](/Users/neeeli01/work/mcp_main/mcp/.github/workflows/black-duck-security-scan-ci.yml) - non-container separate scan.

1. Pull requests receive a rapid source scan
2. Pushes to main and scheduled runs receive a full source scan.
3. Results use Black Duck project/version Arm:MCP / 1.0.
4. BLOCKER and CRITICAL violations fail the workflow.
5. SARIF is uploaded to GitHub Code Scanning, including after policy failures.

**Shared container-scan action** - All container workflows reuse [blackduck-image-scan/action.yml](/Users/neeeli01/work/mcp_main/mcp/.github/actions/blackduck-image-scan/action.yml).
1. Validates the image tag and expected architecture.
2. Checks the locally built image architecture.
3. Exports the image using docker image save.
4. Runs Black Duck with CONTAINER_SCAN & waits for Black Duck results.
5. Uploads SARIF to GitHub Code Scanning & retains it for 10 days
6. Deletes the temporary image archive.
7. Classifies failures: Exit code 8: policy violation & Other failure: scan or integration error.

**MCP build-input images**

Locked dependencies
      ↓
AMD64 local image ──→ Black Duck container scan
ARM64 local image ──→ Black Duck container scan
      ↓
Buildx publication with provenance
      ↓
Compare published configuration to scanned candidate
      ↓
Create multi-architecture index

**Embedding toolchain**

Build ARM64 toolchain locally
      ↓
Black Duck container scan
      ↓
Push the exact scanned local image
      ↓
Generate signed provenance attestation
      ↓
Propose toolchain digest update

**Embedding vectorstore**

Pinned toolchain + intrinsic chunks
      ↓
Generate embedding chunks
      ↓
Build scratch vectorstore image locally
      ↓
Black Duck container scan
      ↓
Push scanned image
      ↓
Attest and verify provenance
      ↓
Propose MCP embedding pin

**Dry-run behavior** - Manual workflows default to dry_run: true.

- Images are built and scanned.
- Black Duck results are uploaded.
- GHCR publication is skipped.
- Provenance and pin-update PRs are skipped.


.github/actions/blackduck-image-scan/action.yml
```diff
@@ -0,0 +1,180 @@
...
+      env:
+        DETECT_PROJECT_NAME: ${{ inputs.project_name }}
+        DETECT_PROJECT_VERSION_NAME: ${{ inputs.project_version }}
+      with:
+        blackducksca_url: ${{ inputs.blackducksca_url }}
+        blackducksca_token: ${{ inputs.blackducksca_token }}
+        blackducksca_scan_full: true
+        blackducksca_scan_failure_severities: 'BLOCKER,CRITICAL'
+        blackducksca_waitForScan: true
+        detect_args: >-
```
brikin01, line 95: Could we set the Black Duck distribution type per image? Most of these images are internal build artifacts that aren’t distributed, while the final runtime image is public.

Black Duck [defaults the project version distribution to `EXTERNAL`](https://docs.blackduck.com/r/detect/latest/black-duck-detect/project.html), and its [license-risk calculation takes the distribution type into account](https://docs.blackduck.com/r/blackduck/latest/black-duck-documentation/determining-license-risk.html).

Could we add a `distribution` input to this action and pass it through here?

```yaml
--detect.project.version.distribution=${{ inputs.distribution }}
--detect.project.version.update=true
```

Based on Black Duck’s [distribution definitions](https://docs.blackduck.com/r/blackduck/2026.4/black-duck-documentation/creating-a-new-version-of-a-project.html), the pipeline-only images could pass `INTERNAL`, while the public MCP runtime would likely pass `OPENSOURCE` because its source is publicly available.

I think `update=true` is needed because these Black Duck version names are reused; otherwise, the distribution is only set when a version is first created.

.github/workflows/trusted-mcp-release.yml
```diff
@@ -287,6 +290,41 @@ jobs:
             --strace-package-version "${STRACE_PACKAGE_VERSION}" \
             --evidence-dir "${EVIDENCE_DIR}"
 
+      - name: Prepare validated runtime image for Black Duck
```
brikin01, line 316: **Non-blocking:** The production architecture tags are published before the Black Duck scan runs, so a failed or timed-out scan can still leave the public `<version>-amd64` and `<version>-arm64` tags on Docker Hub.

This issue is tracked by [STESOL-604](https://jira.arm.com/browse/STESOL-604). We may want to consider addressing that sooner rather than later. Otherwise, this PR blocks the multi-architecture image and final release, but not publication of the architecture-specific tags.

brikin01 (commented):
Thanks for the PR! This looks good in terms of adding Black Duck scanning and release gates, with SBOM work to follow. 

I left one comment about setting the Black Duck distribution type for each artifact, which should ideally reduce the number of license blockers we see. Hopefully adding that doesn't run into issues. 

I also added one non-blocking note about the architecture-specific Docker Hub tags already tracked by [STESOL-604](https://jira.arm.com/browse/STESOL-604). Maybe something I can tackle next week if deemed important.

.github/actions/blackduck-image-scan/action.yml
JoeStech, line 1: Now that this file is going in, please Add /.github/actions/ to CODEOWNERS and extend the ownership test.

.github/workflows/trusted-mcp-release.yml
```diff
@@ -287,6 +290,67 @@ jobs:
...
+          docker logout ghcr.io
+          if [ "${RELEASE_MODE}" = "production" ]; then
+            docker logout
+          fi
+
+      - name: Black Duck scan validated runtime image
+        uses: ./.github/actions/blackduck-image-scan
+        with:
+          project_name: "Arm:MCP"
+          project_version: "mcp-runtime-container-${{ matrix.tag }}-1.0"
```
JoeStech, line 344: runtime scans use the same black duck project version, so overlapping runs can cause a problem, i.e.

  1. Production run A scans release image A into the shared Black Duck version.
  2. Manual dry run B scans a different image into that same version.
  3. Production run A requests the version’s SBOM.
  4. Depending on how Black Duck updates or combines the associated scan data, the exported SBOM can describe B or combined state rather than exclusively A.

I think we should just remove dry-run functionality. dry runs right now don't even exercise a large percentage of the whole pipeline, and maintaining them has been a bit of a pain.

let me know your thoughts here.

.github/scripts/export-blackduck-cyclonedx.py
```diff
@@ -0,0 +1,422 @@
...
+                return
+            if time.monotonic() >= deadline:
+                raise BlackDuckExportError(
+                    "Timed out waiting for Black Duck server-side processing."
+                )
+            time.sleep(poll_interval)
+
+    def _reports(self, version_url: str) -> list[dict[str, Any]]:
+        response = self._json_request(f"{version_url}/reports")
+        items = response.get("items", [])
```
JoeStech: this only gets one response's items and doesn't follow pagination. 

Can you check whether you think we'll hit a scenario where the requested report falls outside the first results page? I don't actually know if this is likely to happen because I don't know black duck

JoeStech (changes requested):

.github/scripts/export-blackduck-cyclonedx.py
NeethuESim: I checked the docs, there is a possibility of results getting stacked up for multiple runs. I will make the changes to read it directly from the url returned by BlackDuck location header.

.github/workflows/trusted-mcp-release.yml
NeethuESim: Technically your concern is valid. However the only reason I added manual dry_run is for testing the blackduck pipeline flow. Copilot is not enough to debug with the logs, need to test out the behaviour eg the timeout issue. Hence added dry_run that doesn't run all the other production flows for every test runs i.e it skips publication, manifest creation, attestations, verification, and GitHub release creation.

As long as we make sure dry_run is manually triggered only for testing we should't run into this overlapping issue. Irrespective of the type of runs, image scanned will be the same, so report generated should be the same. I can also make changes in concurrency to set queue: max which ensures only one production or dry-run workflow runs at a time, with additional runs waiting in order & running workflows are not cancelled.

Nevertheless, I am ok to remove dry_run completely after I test out all the blackduck issues/scenarios 👍 . Like I said, log analysis is not enough for debugging blackduck. So, if we run into issues then might have to bring it back later for debugging.

JoeStech: Thanks, if you find dry run valuable for testing let's keep it in. It's probably overkill to add a new project version just for dry run, since this race condition is not going to happen, practically speaking.

JoeStech (dismissed):

.github/workflows/trusted-mcp-release.yml
NeethuESim: We can fix it after this PR is merged in.

JoeStech (approved):

PR #174, opened 2026-09-10
Reuse BM25 results within hybrid search

sairamvarmabudharaju:
Hybrid search currently scores and sorts the same BM25 query twice: once inside the lexical prepass and again for sparse retrieval. Retrieve one result list at the depth required by both paths, reuse it for lexical reranking, and slice it for sparse retrieval.

Extract the existing lexical ranking logic into a helper so standalone lexical prepass calls keep their behavior. Preserve candidate limits and the existing merging and final ranking rules.

Increase the independent arm-kb-search package version to 0.3.1

### DIFF
--- a/arm_kb_search/search.py
+++ b/arm_kb_search/search.py
@@ -232,6 +232,15 @@ def lexical_prepass_search(
     """Return high-exactness lexical candidates before dense retrieval is merged."""
     prepass_depth = max(k, candidate_depth)
     candidates = bm25_search(query, metadata, bm25_index, prepass_depth)
+    return _rank_lexical_candidates(query, candidates, k)
+
+
+def _rank_lexical_candidates(
+    query: str,
+    candidates: List[Dict[str, Any]],
+    k: int,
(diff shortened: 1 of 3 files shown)


arm_kb_search/search.py
```diff
@@ -232,6 +232,15 @@ def lexical_prepass_search(
     """Return high-exactness lexical candidates before dense retrieval is merged."""
     prepass_depth = max(k, candidate_depth)
     candidates = bm25_search(query, metadata, bm25_index, prepass_depth)
+    return _rank_lexical_candidates(query, candidates, k)
```
brikin01: Thoughts on just removing this function altogether now? I don't think it's called by anything after this change

brikin01 (dismissed):
LGTM, I did leave one comment about possibly removing the now-unused `lexical_prepass_search` function, but left as non-blocking. If you do remove, we also need to remove it from the `init.py`. But nothing that we have is using that function after this rewrite.

brikin01 (approved):

PR #175, opened 2026-09-10
Add production env gate and create action for withdrawing release

brikin01:
- Gate production releases behind approval from the production environment.
- Add a manual workflow to withdraw an MCP release by:
  - restoring latest to a previous version
  - deleting the withdrawn Docker tags
  - deleting the GitHub release and tag
- Add concise withdrawal guidance and test coverage.

### DIFF
--- a/.github/workflows/withdraw-mcp-release.yml
+++ b/.github/workflows/withdraw-mcp-release.yml
@@ -51,6 +51,7 @@ jobs:
         id: validate
         shell: bash
         env:
+          GH_TOKEN: ${{ github.token }}
           WITHDRAW_VERSION: ${{ inputs.withdraw_version }}
           RESTORE_VERSION: ${{ inputs.restore_version }}
           CONFIRMATION: ${{ inputs.confirmation }}
@@ -68,10 +69,20 @@ jobs:
               --format '{{json .Manifest}}' | jq -er '.digest'
           }
 
-          withdraw_digest="$(digest "${IMAGE_FQDN}:${WITHDRAW_VERSION}")"
+          gh release view "v${RESTORE_VERSION}" \
+            --repo "${GITHUB_REPOSITORY}" > /dev/null
+
           restore_digest="$(digest "${IMAGE_FQDN}:${RESTORE_VERSION}")"
           latest_digest="$(digest "${IMAGE_FQDN}:latest")"
-          [ "${latest_digest}" = "${withdraw_digest}" ]
+          withdraw_digest="$(
+            digest "${IMAGE_FQDN}:${WITHDRAW_VERSION}" 2>/dev/null || true
+          )"
+          if [ -n "${withdraw_digest}" ]; then
+            [ "${latest_digest}" = "${withdraw_digest}" ] ||
+              [ "${latest_digest}" = "${restore_digest}" ]
+          else
+            [ "${latest_digest}" = "${restore_digest}" ]
+          fi
 
           echo "restore_digest=${restore_digest}" >> "${GITHUB_OUTPUT}"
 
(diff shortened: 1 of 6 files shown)


mcp-local/tests/test_release_withdrawal.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+REPOSITORY = Path(__file__).resolve().parents[2]
+WORKFLOW = (
+    REPOSITORY / ".github/workflows/withdraw-mcp-release.yml"
+).read_text(encoding="utf-8")
+RUNBOOK = (REPOSITORY / "docs/release-withdrawal.md").read_text(encoding="utf-8")
+
+
+def test_withdrawal_is_manual_guarded_and_serialized() -> None:
```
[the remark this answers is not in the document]
brikin01: Addressed in 58e6ff662538db3f6d8e2bea8077dfc3ca649027

.github/workflows/withdraw-mcp-release.yml
```diff
@@ -0,0 +1,152 @@
...
+
+          digest() {
+            docker buildx imagetools inspect "$1" \
+              --format '{{json .Manifest}}' | jq -er '.digest'
+          }
+
+          withdraw_digest="$(digest "${IMAGE_FQDN}:${WITHDRAW_VERSION}")"
+          restore_digest="$(digest "${IMAGE_FQDN}:${RESTORE_VERSION}")"
+          latest_digest="$(digest "${IMAGE_FQDN}:latest")"
+          [ "${latest_digest}" = "${withdraw_digest}" ]
```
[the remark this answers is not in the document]
brikin01: Addressed in 58e6ff662538db3f6d8e2bea8077dfc3ca649027. Made withdrawals retry-safe by accepting already-completed steps while refusing to overwrite an unexpected latest image.

.github/workflows/withdraw-mcp-release.yml
```diff
@@ -0,0 +1,152 @@
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+          WITHDRAW_VERSION: ${{ inputs.withdraw_version }}
+          RESTORE_VERSION: ${{ inputs.restore_version }}
+        run: |
+          set -euo pipefail
+          gh release edit "v${RESTORE_VERSION}" \
+            --repo "${GITHUB_REPOSITORY}" --latest
+          gh release delete "v${WITHDRAW_VERSION}" \
+            --repo "${GITHUB_REPOSITORY}" --cleanup-tag --yes
```
JoeStech: I don't think we should cleanup tag here. If we keep it then we won't reuse the version, and it gives better history of what actually happened

.github/workflows/withdraw-mcp-release.yml
```diff
@@ -0,0 +1,152 @@
...
+          [ "${WITHDRAW_VERSION}" != "${RESTORE_VERSION}" ]
+          [ "${CONFIRMATION}" = \
+            "WITHDRAW ${WITHDRAW_VERSION} TO ${RESTORE_VERSION}" ]
+
+          digest() {
+            docker buildx imagetools inspect "$1" \
+              --format '{{json .Manifest}}' | jq -er '.digest'
+          }
+
+          withdraw_digest="$(digest "${IMAGE_FQDN}:${WITHDRAW_VERSION}")"
```
JoeStech: gpt-6 comment:

this requires the withdrawn Docker tag to exist and latest to point to it. After “Restore latest” succeeds, any subsequent failure makes those prerequisites false. For example, a Docker deletion permission error leaves cleanup unfinished, but rerunning the workflow immediately fails validation. Deleting the version tag before a later failure also breaks the initial lookup. I reproduced both cases with mocked commands. Allow resuming when latest already matches the fallback digest, tolerate already-removed artifacts, and still reject unrelated latest values.

JoeStech (changes requested):

.github/workflows/withdraw-mcp-release.yml
brikin01: I think this is what copilot also pointed out in an above comment, should hopefully be addressed in the lates version

.github/workflows/withdraw-mcp-release.yml
brikin01: Updated as suggested: the Git tag and GitHub release are retained, the release is marked `WITHDRAWN` with its original notes preserved, and the fallback release becomes latest. Only the withdrawn Docker tags are deleted, and retries won’t duplicate the notice.

JoeStech (approved):

PR #179, opened 2026-09-13
Fix install guide child indexing

jaidev17:
**Summary**
Fixes missing child pages from multi-page Learn install guides.
The chunking pipeline previously indexed only the parent guide. It now discovers and chunks direct child pages while preserving their canonical URLs.
Examples now indexed:
- /install-guides/browsers/chrome/
- /install-guides/docker/docker-engine/
- /install-guides/docker/docker-desktop/

**Testing**
- 63 tests passed
- Ruff passed
- Embedding Docker image built successfully
- Verified against the live Browsers and Docker pages


jaidev17:
Will be fixed through Sai's PR.

PR #180, opened 2026-09-13
Fix install guide ranking

jaidev17:
**Summary**
Improves ranking for queries requesting installation guidance.
Short queries previously received a Learning Path root bonus, while Install Guides had no equivalent preference. This could cause Learning Paths to rank above more relevant Install Guides.

**Changes**
- Detects installation intent from terms such as install, installation, guide, steps, instructions, setup, “set up,” and download.
- Suppresses the Learning Path root bonus for installation queries.
- Adds an Install Guide ranking bonus.
- Supports the production Install Guides document type.
- Preserves Learning Path preference for general usage queries.
- Updates arm-kb-search to 0.3.1.

**Testing**
- Added regression coverage for multiple installation-query variations.
- 59 relevant tests passed.
- Package wheel and dependency lock validated.


jaidev17:
Sai is fixing these issues.

